Issue #32, opened 2026-09-10
workBenches pins openRepoShape and openRepoTools by vendored copy: a multi-source upstream-pin.yaml, an update-upstream.py check|apply, and the estate commands in dev-bench-base (opensoft/openRepoShape#92, W1 + W2)


brettheap:
Lane: xfactory-2 (xFactory-2)

Governing record (lane-collision-protocol Rule 1) for the object **"workBenches pins the versions of
openRepoShape and openRepoTools it uses, and installs their commands into `dev-bench-base`"** —
slices W1 and W2 of opensoft/openRepoShape#92, whose design record covers all three
repositories. This record exists so a lane working in this repository finds the claim in this
repository's own issue space. #30 (the Speckit git extension's `park.sh` / `resume.sh`, realized by
#31) is the layer underneath and is a different, completed object: **nothing in this work touches
`devBenches/base-image/files/speckit-worktree/`**.

## Where this comes from

Brett Heap, 2026-09-10: *"Since we have repo tools like park and resume, i am thinking we carve that
out of openRepoShape and make a new repo openRepoTools. then in workBenches, we pin the both of
these as to the version that we are using in workBenches."*

"Pin" here means what it means in openRepoShape's own grammar: a 40-hex commit plus a sha256 per
file, in a file under review, checked by running code.

## Why a vendored COPY and not a fetch at build time

The Docker build cannot reach github.com reliably: the build network may route it through a VPN
tunnel whose MTU breaks TLS mid-transfer. That fact is already written into
`devBenches/base-image/Dockerfile` next to the Oh My Zsh plugins, which come from apt for exactly
this reason. So the bytes live in this repository, and their identity is the pin.

This is a NEW artifact class here. `git grep` finds no existing convention for "we hold a copy of
somebody else's file at commit X": the sha256 manifests under `.specify/integrations/` are Spec
Kit's own install receipts with a package version and no commit, `.build-state.json` is a
rebuild-trigger digest over local inputs, and the only 40-hex commits in the repository are four
`actions/checkout` pins. The vocabulary is therefore borrowed wholesale from openRepoShape
(`kind: pinned_contract_manifest`, `revision_kind: commit`, `materialization: copied`, `commit:`,
`files: [- path/sha256]`) so that a reader who knows one knows the other, while the reading code is
this repository's own — there is nothing upstream to reuse.

## W1 — vendor openRepoShape, pin it, install the commands

One pull request from `feat/vendor-estate-commands`.

[...]

PR #33, opened 2026-09-10
Vendor the openRepoShape estate commands and pin them: a multi-source upstream-pin.yaml, an update-upstream.py check|apply, openRepoShape/park/resume in /usr/local/bin, and the three-leg fixture derived from the pinned manifest template (#32)

brettheap:
Lane: xfactory-2 (xFactory-2)
Claim: https://github.com/opensoft/workBenches/issues/32#issuecomment-5621452209

Part of #32 (W1 of two slices; W2 follows once opensoft/openRepoTools has a main). Design record: opensoft/openRepoShape#92.

**What lands.** This repository now holds VENDORED COPIES of another repository's commands, and their identity is a file under review. `devBenches/base-image/upstream-pin.yaml` records, per SOURCE repository, a 40-hex commit and a sha256 per file; `devBenches/base-image/update-upstream.py` is the code that says the copies still match it; `dev-bench-base` installs `openRepoShape`, `park` and `resume` into `/usr/local/bin`; and `test-speckit-git-feature.sh` stops hand-mirroring openRepoShape's assembly-root manifest and DERIVES its three-leg fixture from the pinned template.

Copies and not a build-time fetch, for the reason already written into the Dockerfile beside the Oh My Zsh plugins: the build network may route github.com through a VPN tunnel whose MTU breaks TLS mid-transfer. So the bytes live here.

**The pin.** `opensoft/openRepoShape` at `d090a7a141048b214f0420a8c7afbfff62472d1a` — `gh api repos/opensoft/openRepoShape/compare/main...d090a7a1 -q .status` says `identical`, so it is the commit `main` carries, and its tree still holds `park` and `resume`, which is why W1 precedes the carve (S2). It is deliberately post-openRepoShape#93, which rewrote `park` (+288/−90: a bare `park` with no estate now parks every estate) and touched `resume` (+9); pinning the pre-#93 tip would have shipped this image a `park` that main had already replaced. Nothing under `templates/` moved between those two commits (`git diff --stat` empty), so the derived fixture below is unaffected by the move — and that is said from a run, not from reasoning: the whole suite matrix was re-run after the re-vendor.

Four files: `openRepoShape`, `park`, `resume` and `templates/assembly-root/project.yaml`. `vendor_dir` is relative to the pin file's own directory and `path` is BOTH the upstream path and the path under it — one string, two meanings, because the vendored tree mirrors the upstream's layout. There is deliberately no `digests.tree_sha256`, unlike `shape-pin.yaml`: nothing here is mounted, so no offline run could recompute it, and a number no run checks is decoration. Multi-source from day one though W1 fills in one source, so W2 is two `apply` runs and one Dockerfile hunk and changes neither the schema nor the tool.

**The tool.** `check` / `apply` / `list`, python3 standard library only, with its own reader for the subset of YAML the pin is written in — self-contained rather than importing the upstream's `repo_shape.py`, which would give the checker the one property it must not have: breaking on exactly the refactor it exists to notice. Exit 0 ok, 1 a finding, 2 a refusal.

`check` writes NOTHING — no temporary file, no cache, no recomputed row — because it runs inside two CI containers that mount the workspace `:ro`, and it needs no network. It reports a changed byte, a missing file, a copy that stopped being a regular file, and a file under a vendor directory with NO row (every file under a vendor directory is a pinned copy by construction). It never offers to recompute a row: that records the local edit as the upstream, which is the failure `shape-pin.yaml`'s own header spends a paragraph forbidding, so the finding names the two real exits and the actual source id in both of them.

`apply` refuses, and each refusal names its fix: an `--at` that is not a full 40-character lowercase commit (a tag can be moved and a commit cannot); a `--source` the pin does not name (a new source is a new block in that file and a human's decision, not a flag); **a commit not reachable from the source repository's default branch** — openRepoShape squash-merges, so a pull-request commit is orphaned by the merge and a raw fetch of it 404s for the next person, and a pin the rest of the world cannot fetch is a pin nobody can reproduce; a fetch that failed for ANY file in the set, with nothing written, no copy replaced, and every failure named together with both forms it tried (openRepoShape #82's F10 rule); and a run without `--yes`, which prints the per-file plan (`take`/`unchanged`/`add`/`remove`/`re-pin`) and stops. The fetch order is the shims': `gh api …/contents/<path>?ref=<sha>` with `Accept: application/vnd.github.raw` first, because it is authenticated and so survives an organisation that blocks raw.githubusercontent.com, then `raw.githubusercontent.com`.

**The vendored copies came through the tool, not by hand.** `apply --source openreposhape --at d090a7a1… --yes` wrote all four, and they are byte-identical to the upstream tree at that commit: their git blob oids are `d80b6269` / `21990529` / `ee6c5a97` / `a07aeaae` and their modes `100755` / `100755` / `100755` / `100644`, which is exactly what `git ls-tree d090a7a1` reports. `diff <(git show d090a7a1:<f>) devBenches/base-image/files/openreposhape/<f>` is empty for all four, and the sha256 computed from the upstream blob, from the local file and read out of the pin row agree for all four.

**The derived fixture (D7), and it is a strengthening.** `write_three_leg_manifest` substitutes all twenty `{{…}}` placeholders of the pinned `templates/assembly-root/project.yaml` and REFUSES, naming them, if any survives — so a field the standard adds fails here, loudly, instead of leaving a fixture that quietly stops resembling the thing it stands for. Its DATA is byte-identical to the hand-mirror it replaces (diffed with comments and blanks stripped), so every scenario sees the same fields, values and key order. What is new is the template's own ~140-line comment block, which reaches `workspace_read_scalar`, `workspace_project_repository` and `load_repo_shape` for the first time, commented `role:` lines included. **All three handled it, so the risk the plan flagged (a finding in the reader rather than in the fixture) did not materialise, and nothing under `files/speckit-worktree/` is touched by this pull request.** Driven directly against the derived file, `load_repo_shape` returns `REPO_SHAPE=three-leg SHAPE_SPEC_PATH=spec SHAPE_CODE_PATH=code` — the code leg is not mounted at `spec/` — and `workspace_project_repository` returns `dummy/fixture-project`, the assembly leg and not the `shape:` block's repository.

Three properties the helpers depend on are asserted at the derivation: `    repository: dummy/fixture-project` occurs exactly once, so `set_assembly_repository` still works; exactly three legs open with a role at indent 2; and the naive-parser trap survives twice — the code leg's nested `role: spec` is the one a reader that takes the last `role:` it saw would mount at `spec/`. The commented decoy leg is the FIXTURE's own; the template carries none, so it is appended after substitution under a comment saying which lines are the fixture's and why.

**CI.** Four globs in BOTH `paths:` lists (the pin, the tool, and both vendor directories — `files/openrepotools/**` ahead of that directory existing, so W2 touches no workflow). Without them a pull request that only changes vendored bytes fires no job at all and merges green on nothing. One `regression` step runs `check` after the prerequisites and before every suite, so a drifted copy is named in seconds rather than after fifteen minutes of git fixtures. The pre-existing Layer-0 path in this Layer-1a trigger list (`base-image/files/claude/skills/speckit-clarify/SKILL.md`, missing its `devBenches/` prefix) is noticed and left exactly as it is.

**Decisions beyond the plan, for the reviewer.**

1. **`apply` rewrites the rows it owns IN PLACE; it does not re-emit the pin from a schema.** Only the target source's `commit:` line and its `files:` rows change — the header prose, every other source's block, the key order and the trailing bytes survive byte for byte, which is what makes `git diff` after an `apply` readable and is what lets W2's two `apply` runs not disturb each other. The consequence is stated in the pin's own header: a new source block is HAND-WRITTEN once with its `path:` rows and no digests, and the first `apply` fills them in; and a comment placed between two file rows does not survive. The plan's header line "Written by `update-upstream.py apply`; do not hand-edit" is therefore spelled out rather than left absolute.
2. **Modes are not pinned.** A fetched file whose first two bytes are `#!` is placed 0755; anything else 0644. That reproduced the upstream mode for each of these four — but the rule is `#!`, not the upstream bit, and it is lossy: a `.py` that upstream stores 0644 would land 0755 (NIT-3). The pin records BYTES: the Dockerfile chmods 0755 explicitly, so no image depends on the in-tree bit, and a `check` that compared filesystem modes would report a finding on any checkout made by a tool that does not carry them.
3. **The template gets TWO ways to be found, not the suite's three.** The image deliberately does not carry `files/openreposhape/templates/`, so there is no `/usr/local/share` fallback to invent. `devBenches/devcontainer.test/docker-compose.yml` therefore mounts `../base-image` read-only and sets `$WORKBENCHES_SHAPE_TEMPLATE_ROOT` and `$UPDATE_UPSTREAM_FILE` (named for the subsystem as vendored HERE, matching its sibling `$SPECKIT_WORKTREE_TEMPLATE_ROOT` rather than squatting on the upstream product's namespace — NIT-5) — one volume and two variables, which is what keeps `test.sh`'s in-container `/test/test-speckit-git-feature.sh` working when only that directory is the workspace. Without it, a hard fail-fast gate on a file the image does not have would have reddened Layer 1a's own smoke test. This file is not in the plan's file table; it is the cost of the plan's own "NOT copied into the image" decision.
4. **`check` refuses more than a changed byte.** A symlink at a pinned path, a copy that resolves outside its vendor directory (`ESCAPES`), a duplicate `path:` row, two rows where one nests inside the other, a duplicate key in a source block, two `files:` blocks, two top-level `sources:` keys, two `vendor_dir`s where one nests inside the other, an uppercase digest or commit, a row with no digest, and a pin that is not UTF-8 — each exit 2 with the file, the line and the fix. A mis-read pin in a checker is a wrong answer with a confident tone.
5. **A new source's rows are hand-written once, and both refusals say so.** `- path: <upstream path>` per file with no `sha256:` (or an empty one); the first `apply` fetches those paths and fills the digests. `check` refuses an unfilled row (`pin-row-unfilled`) naming the `apply` to run, and `pin-no-rows` names the whole sequence — it is the refusal W2's operator is most likely to meet. `apply --remove` that would empty a source is refused before the fetch, because an empty `files:` block is a pin `check` rejects.

**Verification.** Every number below is from a run on this branch's HEAD.

| Run | Result |
|---|---|
| `update-upstream.py check` | exit 0, four rows named |
| `update-upstream.py list` | exit 0, one source, four rows |
| host `test-speckit-git-feature.sh` (bash 5.2) | **56/56 GREEN**, exit 0 (55 existing + 1 new) |
| host `test-speckit-git-compat.sh` | **22/22 GREEN**, exit 0 |
| host `test-openspeckit-bootstrap.sh` | **742 GREEN**, exit 0 |
| `bash:3.2` + BusyBox, workspace `:ro` (feature + compat) | **78/78 GREEN**, exit 0 (56 + 22) |
| `ubuntu:22.04` git 2.34.1, workspace `:ro` (feature + compat) | **78/78 GREEN**, exit 0 (56 + 22) |
| `test-ct-launchers.sh` with CI's two env vars | exit 0 (silent on success); byte-unchanged by this PR |

Adversarial checks, each run rather than reasoned:

* **Drift.** One byte appended to the vendored `park` → `check` exits 1 naming `CHANGED files/openreposhape/park recorded 432f0023 computed 96a34557` and printing the two exits with the real source id; `git checkout --` → exit 0.
* **Missing.** `resume` moved away → exit 1, `MISSING files/openreposhape/resume`.
* **Unpinned.** A stray file under the vendor directory → exit 1, `UNPINNED files/openreposhape/stray.txt`, and the remediation names `--add`.
* **Malformed pins.** Twenty mutations — truncated, short digest, uppercase digest, duplicate `path:` row, `..` traversal, absolute path, wrong `kind`, wrong `schema_version`, wrong `digest_algorithm`, 39-hex commit, uppercase commit, duplicate source key, duplicate top-level key, two `files:` blocks, `materialization: mounted`, a `source_repository` that is not `owner/repo`, `revision_kind: tag`, a missing `vendor_dir`, a tab in the indentation, an empty `sources:` — each exits **2** and names the file, the line and the fix.
* **`check` writes nothing.** Run over a `chmod -R a-w` copy of the tool, the pin and the vendor directory: the `find -printf '%m %s %T@ %p'` listing digest is identical before and after, and no `__pycache__` or temporary file is left. It also runs green inside the read-only Alpine and Ubuntu mounts above.
* **`apply` is idempotent.** Re-running it at the same commit reports four `unchanged` and leaves the pin byte-identical.
* **`apply --add` / `--remove`** exercised against a scratch pin: `--add setup.sh --remove park --remove resume` took the new file 0755, deleted the two vendored copies, rewrote only the rows, and left lines 1-45 of the pin byte-identical. `--add` of an already-pinned path and `--remove` of an unpinned one both refuse by name.
* **An off-main commit is refused.** `--at 1b3b0e33…` (an openRepoShape branch head, `compare` says `diverged`) exits 2 and writes nothing.
* **A failed fetch writes nothing.** `--add no/such/file --add also-missing` exits 2, names both, prints both attempted forms, and leaves the pin and the vendor directory untouched.
* **Python floor.** `check` runs green on python 3.10 (ubuntu:22.04) and 3.12 (alpine). `flake8` is clean at 100 columns and has 16 E501s at 79 — the same file class as openRepoShape's own tools, which have 24-27 each and no flake8 config.
* **Rule 1.** `git diff origin/main | grep -E '^\+.*(/home/|/Users/|C:\\)'` is empty.
* **Nothing under `devBenches/base-image/files/speckit-worktree/` changed** (`git diff --stat origin/main --` on that path is empty), and no existing scenario in any suite was edited — the suite diff is five hunks: the two locators, the `[ -f ]` gate, the manifest derivation, the new scenario and its one registration line.

**Not proven here: the image.** `./build.sh` was attempted once and failed before reaching any `COPY`, at `resolve image config for docker-image://docker.io/docker/dockerfile:1.7` with `error getting credentials - err: exit status 1` — a registry-credential fault on this workstation, unrelated to this change (no layer of this Dockerfile had run). **The rebuild is P3's step, after W2 merges**, and this is the command that proves it:

```sh
cd devBenches/base-image && ./build.sh
docker run --rm <tag build.sh printed> sh -c \
    'for c in openRepoShape park resume; do command -v "$c" || exit 1; done
     find / -name project.yaml -path "*openreposhape*" 2>/dev/null'
```

Three `/usr/local/bin` paths, and NO `project.yaml` from the vendored templates anywhere in the image — the Dockerfile copies the three commands and not `files/openreposhape/templates/`. **NIT-6, worth writing down because the plan and this body both had it wrong:** under POSIX `sh`, `command -v a b c` proves only the FIRST name and prints one line, so the loop above (or `bash -c`) is the form that actually proves three. The adversarial review built the image from this branch and ran the strong form: three `/usr/local/bin` paths, modes 0755, the three `sha256sum`s equal to the pin rows, and `find / -path '*openreposhape*'` empty — so the criterion is met even though this PR does not build it.

**One operational consequence for P3 and for anyone running `test.sh` (NOTE-2):** an already-running `devbench-test` container must be recreated (`docker compose down && docker compose up -d`) to pick up the new `:ro` mount and the two variables, or the new fail-fast gate makes Layer 1a's smoke test lines 93/94 go red. The gate is deliberate and names the fix; the compose change supplies it. A change to `docker-compose.yml` alone fires no CI job, which is harmless because no job reads that file (`regression` runs the suites directly and the two container jobs use `docker run --volume …:ro`).

**Review comments.** The adversarial review returned **0 defects / 8 nits / READY**; the Copilot review flagged six inline findings plus eight suppressed ones. This commit fixes NIT-1, NIT-2, NIT-4, NIT-5 and NIT-7, corrects decision 2's wording (NIT-3), and takes NIT-6 into the image-proof command above. From Copilot it also fixes: a second top-level `sources:` block silently merging into the first; `rewrite_pin` dropping the `- ` marker when `commit:` opens a source item; a repeated `--add`/`--remove` value writing duplicate rows; an undecodable pin producing a traceback instead of a named exit 2; `--remove` leaving a directory or a broken link behind its deleted row; and both verbs following a symlinked `vendor_dir`, a symlinked component or a symlinked destination — `check` now reports `ESCAPES` and `apply` refuses to write through a link, proven by pointing a destination at a path outside the tree and watching nothing be created there. It also strengthens the fixture's leg assertion to name `assembly`/`spec`/`code` once each rather than count three list items, allows a legitimately empty upstream file to be vendored (the exit code is the check, not the body length), and refuses two rows or two `vendor_dir`s where one nests inside the other (NOTE-5). NIT-8 is left as the review recommended: `TOOL` is a literal, correct for every real invocation.

**The one class of comment not acted on here** is Copilot's findings inside `devBenches/base-image/files/openreposhape/resume` (`:724`, `:1190`, `:1283` and their suppressed twins about `ROOT_REFUSED`): those are openRepoShape's own bytes at the pinned commit, and editing a vendored copy is precisely the drift this pin exists to forbid — `update-upstream.py check` would go red on the next run and the recorded fix is to carry the change upstream, never to edit the copy — so they belong in opensoft/openRepoShape, not in this pull request.

**Not here.** openRepoTools' own content (T1); the removal of `park` and `resume` from openRepoShape (S2); the second source and the four-command Dockerfile hunk (W2); the base-image rebuild and the host reinstall (P3, operator). The `re-run \`make park\`` messages in `park.sh` and `resume.sh` are not wrong and are not touched — those scripts ship INTO user projects, where openRepoShape's assembly-root Makefile supplies that target.

## Summary by Sourcery

Vendor and validate pinned openRepoShape commands while deriving the three-leg test fixture from the pinned upstream manifest template.

New Features:
- Vendor the pinned openRepoShape estate commands and assembly-root manifest template for use by the base image and test fixtures.
- Add an offline upstream pin management tool with check, apply, and list workflows for validating and updating vendored copies.
- Derive the three-leg test manifest from the pinned upstream template and validate its required structure.

Bug Fixes:
- Prevent CI and read-only test environments from silently using drifted or untracked vendored upstream files.
- Ensure upstream updates refuse unreachable commits, incomplete fetches, malformed pins, and unsafe filesystem destinations without partial writes.

Enhancements:
- Install openRepoShape, park, and resume into the base image while keeping the fixture template out of the image.
- Run vendored-copy validation in the regression workflow and feature test suite.
- Expose the vendored fixture inputs to the devcontainer test environment through a read-only mount and dedicated environment variables.

CI:
- Trigger the Bash workflow when the upstream pin, updater, or vendored upstream directories change.
- Add a regression step that verifies vendored files match their recorded pins.

Tests:
- Replace the hand-maintained three-leg manifest fixture with a derived fixture based on the pinned upstream template.
- Add coverage for vendored-copy pin validation within the feature test suite.

### DIFF
--- a/.github/workflows/speckit-git-bash.yml
+++ b/.github/workflows/speckit-git-bash.yml
@@ -13,6 +13,10 @@ on:
       - devBenches/base-image/files/claude/commands/opsx/**
       - base-image/files/claude/skills/speckit-clarify/SKILL.md
       - devBenches/base-image/files/ct/ct-functions.zsh
+      - devBenches/base-image/upstream-pin.yaml
+      - devBenches/base-image/update-upstream.py
+      - devBenches/base-image/files/openreposhape/**
+      - devBenches/base-image/files/openrepotools/**
   pull_request:
     paths:
       - .github/workflows/speckit-git-bash.yml
@@ -25,6 +29,10 @@ on:
       - devBenches/base-image/files/claude/commands/opsx/**
       - base-image/files/claude/skills/speckit-clarify/SKILL.md
       - devBenches/base-image/files/ct/ct-functions.zsh
+      - devBenches/base-image/upstream-pin.yaml
+      - devBenches/base-image/update-upstream.py
+      - devBenches/base-image/files/openreposhape/**
+      - devBenches/base-image/files/openrepotools/**
   workflow_dispatch:
 
 permissions:
@@ -55,6 +63,9 @@ jobs:
           bash --version | sed -n '1p'
           python3 --version
           zsh --version
+
+      - name: Check the vendored upstream copies against their pin
+        run: python3 ./devBenches/base-image/update-upstream.py check
 
       - name: Run bootstrap suite
         run: bash ./devBenches/devcontainer.test/test-openspeckit-bootstrap.sh
--- a/devBenches/base-image/Dockerfile
+++ b/devBenches/base-image/Dockerfile
@@ -135,6 +135,33 @@ RUN chmod 0755 /usr/local/bin/speckit-worktree-enable && \
     find /usr/local/share/speckit-worktree/templates -type f ! \( -name '*.sh' -o -name '*.zsh' -o -name '*.ps1' \) -exec chmod 0644 {} +
 
 # ========================================
+# ESTATE COMMANDS (openRepoShape)
+# ========================================
+# VENDORED, never fetched at build time: the build network may route github.com
+# through a VPN tunnel whose MTU breaks TLS, which is why the Oh My Zsh plugins
+# in Layer 0 come from apt rather than a clone. These copies are pinned by
+# commit and per-file sha256 in ../upstream-pin.yaml; `update-upstream.py check`
+# is what says they still are, and it runs in CI (.github/workflows/
+# speckit-git-bash.yml) and in the devcontainer suite.
+#
+# files/openreposhape/templates/ is NOT copied. It is a fixture input for
+# devBenches/devcontainer.test/test-speckit-git-feature.sh, which derives its
+# three-leg project.yaml from the pinned template; the image installs commands.
+# Do not "fix" the omission.
+#
+# PATH, said out loud: Layer 0 writes $HOME/.local/bin AHEAD of the inherited
+# PATH into /etc/skel/.zshrc and /etc/skel/.bashrc (base-image/Dockerfile:261
+# and :270), so `openRepoShape --install` run in an interactive container shell
+# SHADOWS these pinned copies with whatever the upstream main holds. A
+# non-interactive `docker exec` uses this image's own ENV PATH, where these
+# win. Accepted, not fixed: fetching at build time is what the network
+# constraint above forbids.
+COPY files/openreposhape/openRepoShape /usr/local/bin/openRepoShape
+COPY files/openreposhape/park /usr/local/bin/park
+COPY files/openreposhape/resume /usr/local/bin/resume
+RUN chmod 0755 /usr/local/bin/openRepoShape /usr/local/bin/park /usr/local/bin/resume
+
+# ========================================
 # AI META-HARNESS (OMNIGENT)
 # ========================================
 # Omnigent wraps the Layer 0 codex/claude CLIs behind one CLI + web UI.
--- a/devBenches/base-image/Dockerfile
+++ b/devBenches/base-image/Dockerfile
@@ -156,6 +156,11 @@ RUN chmod 0755 /usr/local/bin/speckit-worktree-enable && \
 # non-interactive `docker exec` uses this image's own ENV PATH, where these
 # win. Accepted, not fixed: fetching at build time is what the network
 # constraint above forbids.
+#
+# `openRepoShape --version` prints `(opensoft/openRepoShape @ main)`, not the
+# pinned commit: $OPENREPOSHAPE_REF defaults to `main` upstream and this image
+# does not set it. The identity of these bytes is ../upstream-pin.yaml, not
+# what the command says about itself.
 COPY files/openreposhape/openRepoShape /usr/local/bin/openRepoShape
 COPY files/openreposhape/park /usr/local/bin/park
 COPY files/openreposhape/resume /usr/local/bin/resume
--- a/devBenches/base-image/files/openreposhape/openRepoShape
+++ b/devBenches/base-image/files/openreposhape/openRepoShape
@@ -1,0 +1,310 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# openRepoShape — the short way in: `openRepoShape Atlas --org <your-org>`.
+#
+# Install it once, from anywhere, with no checkout and no clone:
+#
+#   gh api repos/opensoft/openRepoShape/contents/openRepoShape \
+#       -H 'Accept: application/vnd.github.raw' | bash -s -- --install
+#
+# (or the raw-URL twin, `curl -fsSL …/main/openRepoShape | bash -s -- --install`).
+#
+# IT ADDS NO BEHAVIOUR: it fetches `setup.sh` from opensoft/openRepoShape and
+# runs it, turning the positional <Project> into `--project <Project>` and
+# passing every other argument through untouched. The README's long
+# `curl … | bash -s -- --org … --project …` line and this do the same thing.
+#
+# `--install` PLACES THREE COMMANDS, not one: this file, `park` and `resume`
+# (#82). One install line for the whole standard, so a second engineer gets
+# the estate verbs from the README's one line and nothing depends on anyone's
+# dotfiles. The two of them add no behaviour either — they find the estate and
+# run its own `make park` / `make resume`.
+
+set -euo pipefail
+
+# $OPENREPOSHAPE_REPO is an owner/name slug here (it is an API path), while
+# setup.sh reads the same variable as something to `git clone` — see
+# shape_clone_url below: one name, both uses.
+REPO="${OPENREPOSHAPE_REPO:-opensoft/openRepoShape}"
+REF="${OPENREPOSHAPE_REF:-main}"
+SELF="${BASH_SOURCE[0]:-$0}"
+
+PROJECT=""
+ORG=""
+MODE="run"
+DOCTOR=0
+PASSTHROUGH=()
+WORKDIR=""
+STAGED=""
+
+say() { printf '%s\n' "$*"; }
+die() { printf '\nREFUSED: %s\n' "$1" >&2; exit "${2:-2}"; }
+
+usage() {
+	cat <<'USAGE'
+openRepoShape <Project> [--org <org>] [setup-project.py options] [-- <scaffold flags>]
+openRepoShape --install            install (or update) this command into ~/.local/bin
+openRepoShape --doctor             check this machine and stop; creates nothing
+openRepoShape --help | --version
+
+<Project> is the assembly-root name, ONE CamelCase token (e.g. Atlas); omit it
+and setup.sh asks, if a terminal is attached. `--org` comes from that flag,
+else $OPENREPOSHAPE_ORG, else a prompt — it is NEVER defaulted. Every other
+argument reaches setup.sh unchanged: --visibility, --elected-by, --family,
+--into, --yes, --local-remote-dir, and anything after `--`, which setup.sh
+hands on to scaffold-project.py. `./setup.sh --help` prints setup-project.py's
+usage, which lists them all.
+
+`--install` places THREE commands, and each is one file:
+
+  openRepoShape <Project> --org <org>    scaffold a project
+  park [<Name>]                          commit, push and RECORD an estate's
+                                         open features, so another workstation
+                                         can take the work up
+  resume [<Name>]                        clone or fast-forward that estate here
+                                         and recreate those features
+
+`park --help` and `resume --help` say the rest. Neither implements any of the
+mechanics: they find the estate under your projects directory and run its own
+`make park` / `make resume`, which run the Speckit git extension's scripts.
+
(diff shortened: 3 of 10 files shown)


brettheap:
@codex review

PR #36, opened 2026-09-10
update-upstream.py: a row's mode comes from upstream's git tree and check verifies it; TOOL is the tool's own path (#32/#33 NIT-3, NIT-8)

### LINKED ISSUE #35: update-upstream.py: a row's mode comes from upstream's git tree and check verifies the working-tree executable bit; TOOL is the tool's own path, not a hardcoded literal (#32/#33 NIT-3, NIT-8)

Follow-up to #32 (closed) and PR #33: two review findings recorded in #33's
body as "left, not fixed" are now to be fixed.

**NIT-3 — modes.** `update-upstream.py apply` currently sets a fetched
file's mode from its own `#!` bytes (0755 if it starts `#!`, 0644
otherwise); upstream's real mode can differ — `scripts/repo_shape.py` is
`100644` upstream in opensoft/openRepoShape and the `#!` heuristic would
still place it 0755. Fix: take the mode from the source repository's own git
tree at the pinned commit (the trees API,
`gh api repos/<slug>/git/trees/<commit>?recursive=1`, gives `mode` per path;
one call per source, cached for the run) and record it as a `mode:` field on
each row (`"100755"`/`"100644"`), so `check` can verify the working-tree
executable bit against the row — a mismatch is drift, exit 1, naming the
file and the fix (`chmod`). Existing rows without `mode:` keep validating
(treat as unrecorded, no check), so the pin stays readable across the
change; `apply` writes `mode:` for every row it touches. Fetch order and
all-in-hand-before-placing are unchanged.

**NIT-8 — the `TOOL` literal.** Remediation lines print a hardcoded path
(`devBenches/base-image/update-upstream.py`) for the tool. Fix: derive it
from the tool's own path relative to the repository root (`__file__` and
`git rev-parse --show-toplevel`), falling back to the literal when not in a
checkout — so the printed command is right from any cwd and in a container
mount.

No re-vendor, no pin-commit change: a re-`apply` at the same commits is
allowed only to backfill the new `mode:` fields, and the vendored bytes must
stay identical.

brettheap:
Claim: https://github.com/opensoft/workBenches/issues/35#issuecomment-5624658807

Closes #35. Follow-up to #32 / PR #33: fixes the two review findings that PR #33's body recorded as "left, not fixed" (NIT-3, NIT-8).

**NIT-3 — modes now come from upstream, not from a `#!` guess.** `apply` previously set a fetched file's mode from its own first two bytes (`#!` → 0755, else 0644); upstream's real mode can differ (a `.py` upstream ships 0644 would still land 0755 under the old rule). `apply` now reads every path's own mode from the source repository's git tree at the pinned commit — one `gh api repos/<slug>/git/trees/<commit>?recursive=1` call per source, cached for the run — and records it as an optional `mode:` field on the row (`"100755"` / `"100644"`). `check` compares the working tree's executable bit against a row's `mode:` **only when the row has one**; a mismatch is drift, exit 1, naming the file and the fix (`chmod <octal> <path>`). A row with no `mode:` (written before this existed, or hand-written once for a new source with no digest yet) keeps validating exactly as before — unrecorded, not refused — so the pin stays readable across the change. `apply` writes `mode:` for every row it touches, since it always re-fetches and re-pins every row of the source it is given. Fetch order and all-in-hand-before-placing are unchanged; a new `upstream-mode-unusable` refusal (all-or-nothing, before any write) covers a wanted path whose upstream tree entry is not an ordinary file or a file whose executable bit git tracks (a symlink, a submodule, a directory).

**NIT-8 — `TOOL` is the tool's own path, not a literal.** Every remediation line printed a hardcoded `devBenches/base-image/update-upstream.py`, which was wrong whenever the tool or the pin lived elsewhere (a reviewer's throwaway copy, a worktree, a differently-named clone). `TOOL` is now derived from `__file__` and `git rev-parse --show-toplevel` (run at this file's own directory, independent of the caller's cwd), falling back to the old literal when there is no checkout to ask. Verified against three cases: the real checkout (unchanged answer), a copy placed at a different relative path inside a *different* git repository (answer follows the new path), and a copy outside any git repository at all (falls back to the literal).

**The pin's bytes did not move.** Both sources were re-`apply`'d at their *existing* pinned commits (`0fa73b7d…` for `openreposhape`, `9fbf01c7…` for `openrepotools`) purely to backfill the new `mode:` field — `git diff --stat -- devBenches/base-image/files/` is empty, and `upstream-pin.yaml`'s diff is five new `mode:` lines and nothing else (no commit, no sha256 touched). The upstream modes recorded (0755/0755/0755/0755/0644) match what was already committed for these five files, so no vendored copy's actual permission bits changed either.

**Verification.**

| Run | Result |
|---|---|
| `update-upstream.py check` (before backfill) | exit 0, 5 rows match — no `mode:` yet, so no mode check fires |
| `update-upstream.py apply --source openreposhape --at <pinned> --yes` | exit 0, 2 `unchanged`, modes 0755/0644, `git diff --stat -- files/` empty |
| `update-upstream.py apply --source openrepotools --at <pinned> --yes` | exit 0, 3 `unchanged`, modes 0755/0755/0755, `git diff --stat -- files/` empty |
| `update-upstream.py check` (after backfill) | exit 0, 5 rows match, mode compared for all 5 |
| `update-upstream.py list` | exit 0 |
| `chmod 0644` a vendored 0755 file, then `check` | exit 1, `MODE files/openrepotools/park recorded 0755 working tree 0644`, fix line `chmod 0755 files/openrepotools/park`; reverted, `check` exit 0 again |
| Scratch pin: invalid `mode: "100777"` | exit 2, `pin-row-mode`, names file/line/value |
| Scratch pin: `mode:` set twice on one row | exit 2, `pin-duplicate-key` |
| Scratch pin: no `mode:` row at all, file executable | exit 0 — unrecorded rows are not checked |
| host `test-speckit-git-feature.sh` | **56 PASS / 0 FAIL**, exit 0, GREEN — same count as #33 |
| host `test-speckit-git-compat.sh` | **22 PASS / 0 FAIL**, exit 0, GREEN — same count as #33 |
| host `test-openspeckit-bootstrap.sh` | **742 GREEN**, exit 0 — same count as #33 |
| `flake8 --max-line-length=100` | clean, exit 0 |
| Rule 1 (`git diff origin/main | grep -E '^\+.*(/home/|/Users/|C:\\)'`) | empty |

**Not here.** Every other NIT and NOTE recorded in #33's review (NIT-1/2/4/5/6/7, NOTE-1..5) — those were already fixed or dispositioned before #33 merged. This PR touches only `devBenches/base-image/update-upstream.py` and `devBenches/base-image/upstream-pin.yaml`; no vendored byte, no pin commit, no CI workflow, no test suite changed.

## Summary by Sourcery

Pin vendored file modes from upstream Git trees and validate permission drift with accurate tool paths.

New Features:
- Record each vendored file's upstream Git executable mode in the pin and validate recorded modes during checks.
- Derive remediation commands from the tool's actual repository-relative path, with a fallback for files outside Git checkouts.

Bug Fixes:
- Prevent mode drift between upstream files and vendored copies by replacing the byte-content shebang heuristic with upstream tree metadata.
- Reject upstream paths that are not regular files with supported Git modes before writing any files.

Enhancements:
- Preserve compatibility for pin rows without a recorded mode while applying and re-pinning all touched rows with upstream mode metadata.

Documentation:
- Update the tool's documented mode and check behavior to describe upstream mode pinning and optional mode validation.

Chores:
- Backfill mode metadata for the existing pinned upstream files without changing their bytes, digests, or pinned commits.

### DIFF
--- a/devBenches/base-image/update-upstream.py
+++ b/devBenches/base-image/update-upstream.py
@@ -58,6 +58,10 @@ WHAT `apply` REFUSES, because a pin it cannot justify must not be recorded:
     vendored copy is replaced, and every file it could not get is named, with
     both forms it tried. All the bytes in hand before any one of them is
     placed — openRepoShape #82's F10 rule, for its reason;
+  * a path whose mode, in the source repository's tree at that commit, is not
+    an ordinary file's or one whose executable bit git tracks — a symlink, a
+    submodule, a directory. A pin records copies of regular files, and a row
+    naming one of those is not one;
   * a `--remove` set that would leave a source with NO rows. An empty `files:`
     block is a pin this tool's own `check` refuses, so `apply` must not be able
     to write one and then print `NEXT … check`; emptying a source means
@@ -72,12 +76,23 @@ raw.githubusercontent.com; `curl` at the raw URL second, for a machine with no
 `gh`. Each attempt is written to a temporary file and checked before its bytes
 are used, so a half-written first attempt is never mistaken for a fetch.
 
-MODES ARE NOT PINNED, AND THAT IS DELIBERATE. A fetched file whose first two
-bytes are `#!` is a command and is placed 0755; anything else is data and is
-placed 0644. The pin records BYTES: the Dockerfile chmods 0755 explicitly when
-it installs the commands, so no image depends on the in-tree bit, and a
-`check` that compared filesystem modes would report a finding on any checkout
-made by a tool that does not carry them.
+MODES ARE PINNED FROM UPSTREAM, NOT GUESSED FROM BYTES. `apply` reads each
+file's own mode from the source repository's git tree at the pinned commit —
+one `git/trees/<commit>?recursive=1` call per source, cached for the run —
+because a byte-content heuristic is lossy: a `.py` upstream ships 0644 does
+not become a command by starting with `#!`, and nothing about the bytes says
+whether upstream tracks the executable bit at all. The mode is recorded as an
+OPTIONAL `mode:` field on a row, `"100755"` or `"100644"`; `apply` writes it
+for every row it touches. A row written before this existed, or written by
+hand for a new source, carries no `mode:` and stays valid — UNRECORDED, not
+unchecked because broken — so the pin stays readable across the change.
+
+`check` compares the working tree's own executable bit against a row's
+`mode:` ONLY when the row has one; a mismatch is drift, exactly like a
+changed byte, and the fix it names is `chmod`. The Dockerfile still chmods
+0755 explicitly when it installs the commands into the image, unaffected by
+any of this: that bit is about the image, and this one is about the copy
+sitting in this tree.
 
 STANDARD LIBRARY ONLY, and its own reader for the small subset of YAML the pin
 is written in. Importing the upstream's `repo_shape.py` would give this checker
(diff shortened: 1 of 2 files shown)


brettheap:
@codex review

PR #39, opened 2026-09-10
devBenches/base-image: re-vendor openRepoTools at ff1f01a — park skips roots without the overlay, --help wording fix

brettheap:
Lane: openreposhape-2 (openRepoShape-2)
Claim: https://github.com/opensoft/workBenches/issues/38#issuecomment-5625616302

Closes #38.

**What lands.** `devBenches/base-image/upstream-pin.yaml`'s `openrepotools` source moves from `9fbf01c789c311aa7c190b238a3df48ea39473e1` (pinned in #32/#34) to `ff1f01aae70b8f21dce0e4654da071a1281b41a4` — `gh api repos/opensoft/openRepoTools/compare/main...ff1f01aae70b8f21dce0e4654da071a1281b41a4 -q .status` says `identical`, so it is the commit `main` carries. **The tool and the Dockerfile do not change** — the vendored paths are identical to W2's, so this is one `apply` run and nothing else.

**Upstream, between the two commits:**

- opensoft/openRepoTools#7 (closes opensoft/openRepoTools#6): `park`'s bare no-estate sweep SKIPS a standalone root with no Speckit git overlay instead of counting it as a refusal — Brett Heap's ruling on opensoft/openRepoShape#92's open item. `park <Name>` on such a root still refuses; `resume` is untouched.
- opensoft/openRepoTools#4 (closes opensoft/openRepoTools#3): `openRepoTools --help`'s `usage()` sentence naming the two verbs `--install` places alongside it is reworded — wording only.

**The `apply` run, verbatim:**

```
$ python3 devBenches/base-image/update-upstream.py apply --source openrepotools --at ff1f01aae70b8f21dce0e4654da071a1281b41a4 --yes
source      openrepotools (opensoft/openRepoTools)
pinned      9fbf01c789c3
target      ff1f01aae70b (on main)
vendor      files/openrepotools

  take      openRepoTools  15e4cb8ca6bb
  take      park  6c2dca99f2e1
  unchanged resume
  re-pin    commit 9fbf01c789c3 -> ff1f01aae70b

  wrote     files/openrepotools/openRepoTools  0755
  wrote     files/openrepotools/park  0755
  wrote     files/openrepotools/resume  0755
  re-pinned devBenches/base-image/upstream-pin.yaml  3 row(s) at ff1f01aae70b

NEXT  python3 devBenches/base-image/update-upstream.py check
```

[...]

Issue #37, opened 2026-09-10
setup.sh installs the estate commands on the host from the vendored pin — openRepoShape, openRepoTools, park, resume placed from files/openreposhape and files/openrepotools, re-placed when the pin moves


brettheap:
Lane: xfactory-2 (xFactory-2)

Governing record (lane-collision-protocol Rule 1) for the object **"workBenches `setup.sh` installs the estate commands on the host from the vendored pin, and re-places them when the pin moves"**. Sequel to #32 (W1/W2: the pin and the vendored copies, landed #33 → f4b943d, #34 → 3eae325). This work reads that pin; it does not change it.

## Where this comes from

Brett Heap, 2026-09-10, after confirming that nobody clones openRepoShape and that the two installers do not chain (`openRepoShape --install` places one file and prints a pointer at openRepoTools' installer; `openRepoTools --install` places three and never the shim): *"why not make this part of workBenches. so when I run setup in workBenches it checks the versions of them and reinstalls if needed."*

## What is already true, and why this is small

- `devBenches/base-image/upstream-pin.yaml` pins both standards by commit and per-file sha256 (`openreposhape` at 0fa73b7…, `openrepotools` at 9fbf01c…), and `files/openreposhape/openRepoShape`, `files/openrepotools/{openRepoTools,park,resume}` are the vendored bytes the image gets. `update-upstream.py check` proves, offline, that those bytes are the pinned ones.
- Both shims' `--install`, **run from a file**, install those bytes with no network: each target is compared with `cmp` and reported `already installed (unchanged)`, `updated`, or `installed`; all-or-nothing; `$OPENREPOSHAPE_BIN_DIR` / `$OPENREPOTOOLS_BIN_DIR` (default `~/.local/bin`) say where.
- Today the Dockerfile copies the four files into the image and the host is left to two `curl … | bash -s -- --install` lines that fetch `main`. So a bench and its host can run different commands, and the host's copy is whatever `main` was that day.

So the step is: check the vendor dirs, then run the two vendored installers. No version logic of its own.

## Rulings (Rule 2; ANY lane executes under the Rule 1 claim below)

1. **"Version" means the pin, not a number.** The shims' `--version` prints `@ main`, never the pinned commit, so the check is the byte compare the installers already do, against the vendored copies at the commits `upstream-pin.yaml` names. The step prints those two commits so the output reads as a version. Nothing in this work adds a version string to either shim.
[...]

PR #40, opened 2026-09-10
setup.sh installs the estate commands on the host from the vendored pin: openRepoShape, openRepoTools, park, resume placed from files/openreposhape and files/openrepotools, re-placed when the pin moves (#37)

brettheap:
Lane: xfactory-2 (xFactory-2)
Claim: https://github.com/opensoft/workBenches/issues/37#issuecomment-5625482029
Governing issue: #37

## What this does

Adds `scripts/setup-estate-commands.sh`, which installs `openRepoShape`, `openRepoTools`, `park` and `resume` on the host from workBenches' own vendored pin (`devBenches/base-image/upstream-pin.yaml`). It runs `update-upstream.py check` first; only if that passes does it print the two pinned commits and run the two vendored shims' own `--install`, invoked as files so each places its own bytes (and `openRepoTools` copies `park`/`resume` from beside itself) with no network and no version logic of this script's own. `setup.sh` calls it under a new `log_header "ESTATE COMMANDS"` block, after shell setup and before the interactive selector, best-effort like the Wave widgets step (`WORKBENCHES_SKIP_ESTATE_COMMANDS=1` skips it; any failure here is caught and setup continues). The regression test is wired into `.github/workflows/speckit-git-bash.yml` immediately after the existing pin-check step. A paragraph under `## setup.sh Flow`, plus a sentence under `### Re-running setup.sh`, documents the mechanism in the README.

**Review round:** an adversarial review found and named six reproducible defects, D1-D6; all six are fixed in `a4bdbe8`, entirely inside the script, its test, and the README — the vendored shims are untouched (ruling 4).

## The four rulings, and where each lives in the diff

1. **"Version" means the pin, not a number.** Reads the two pinned commits from `update-upstream.py list` (a stable, parseable form) and prints `openRepoShape pinned at <40hex>` / `openRepoTools pinned at <40hex>`. Neither shim gains a `--version` change.
2. **The host follows the pin, both ways.** Invokes both shims' `--install` unchanged, relaying their own per-file `already installed ... (unchanged)` / `updated` / `installed` verbs. `setup.sh`'s new block is default-on and best-effort, matching the Wave widgets pattern.
3. **Drift refuses.** `update-upstream.py check` runs before anything is placed; its exit code is now read precisely: 1 (drift) prints the existing refusal and exits 1; any other non-zero (a structural pin refusal) gets its own message and exits 2 (fix for D5).
4. **Nothing upstream changes.** No commit to openRepoShape/openRepoTools, and this diff does not touch the Dockerfile, `upstream-pin.yaml`, `files/`, or `files/speckit-worktree/`.

## Guarantees added by the review round (D1-D6)

- **D1 (symlinked target)** — the shim's `cp` wrote through a symlinked `park`, silently editing another repository. **Fixed:** pre-flight refuses (exit 1) before either shim runs if any of the four targets is a symlink, naming it.
- **D2 (directory-named target)** — `cp` copied *into* a directory standing in for a target, reporting success while installing nothing. **Fixed:** pre-flight refuses if a target exists and is not a regular file.
- **D3 (unpinned bytes reported as the pin)** — removing a pin row the documented way (`apply --remove`, which deletes the row and the file together) let `check` pass while the shim silently fetched the missing file from the network at `$*_REF`/`$*_REPO`. **Fixed two ways:** (a) pre-flight now also requires the pin file to exist, every file either shim needs to be present on disk, and every file the shims would install to have a row in `update-upstream.py list`'s own output — refusing (exit 2) and naming the file before any shim could reach for the network; (b) for defense in depth, `OPENREPOSHAPE_REPO`/`_REF` and `OPENREPOTOOLS_REPO`/`_REF` are overridden to a sentinel (`pinned-by-workBenches-no-fetch`) that cannot resolve, for the duration of both installer runs, so a fetch that ever became reachable despite (a) fails loudly naming the sentinel instead of silently succeeding against the operator's own exported environment.
- **D4 (partial installs, two independent transactions)** — a read-only pre-existing `park` let `openRepoShape`/`openRepoTools` install before the second shim failed partway through its own three-file loop, leaving a mismatched trio. **Fixed:** the same pre-flight checks all four targets (existence, symlink, regular-file, writability, and bin-dir writability) before *either* shim runs, so a bad target refuses before anything at all is placed.
- **D5 (`check` exit 2 misreported as drift)** — see ruling 3 above; the pin file's own existence is also now checked explicitly before `check` even runs.
- **D6 (hygiene)** — removed this script's own third "not on `$PATH`" line (both shims already print their own) and the `seen`-dedupe it needed; switched both `list`-output parses from a `printf | awk '{exit}'` pipeline to a here-string, so there is no live producer process an early `awk exit` could ever send SIGPIPE under `pipefail` if the output ever grew.

After both installers exit 0, all four placed files are independently re-verified — regular, non-symlink, mode exactly 0755, `cmp`-identical to the vendored copy — and only then does the script report `Estate commands verified against the vendored pin.` (closes D1-D4 residue in one place).

## Tests

`devcontainer.test/test-setup-estate-commands.sh`, **76 assertions, all green**:
```
GREEN: setup-estate-commands regression test passed
```
Original (a)-(e) unchanged in behavior (scenario (a) now also asserts exactly two PATH warnings — one per shim, not a third from this script). New:
- **(g) [D1]** `park` a symlink into another temp dir → exit 1, refusal names it, the link target's bytes and the rest of the bin dir untouched.
- **(h) [D2]** `park` a directory → exit 1, refusal, nothing copied into it.
- **(i) [D4 guard]** pre-seeded read-only `park` (0444) → exit 1 *before* `openRepoShape`/`openRepoTools`/`resume` are placed.
- **(j)** a base-image copy with `upstream-pin.yaml` deleted → exit 2, "missing", bin dir empty.
- **(k) [D3]** `resume`'s row and vendored file both removed via `apply --remove` (confirmed `check` still passes first) → our pre-flight exits 2 naming `resume`, bin dir empty, no "fetch"/"github" in the output.
- **(l)** every file placed in scenario (a) is mode 0755 (`stat -c %a`).
- **(m)** the sentinel: run with operator-style `*_REF`/`*_REPO` values set → still exit 0, unchanged install, neither operator value reaches the output. Since neither shim echoes `$REPO`/`$REF` on the happy path (only on `--version` or inside an unreachable fetch-failure message), a live grep-for-zero-occurrences can't by itself distinguish "overridden" from "never consulted" — so this scenario also asserts, by reading the script's own source, that all four `export …="$ESTATE_SENTINEL"` lines are present, per the reviewer's own fallback ("assert instead by a method you can justify, or drop (m) and say so").

`devcontainer.test/test.sh` is a Layer 0 *container* harness (registers only 3 of its 10 sibling `test-*.sh` files, as in-container helpers) — not a general runner list. `test-ai-cli-install-summary.sh`, host-side like this one, is already unregistered there, so this test follows that precedent and is not added to it; it is wired into CI directly.

Also: `bash -n` clean on both scripts and `setup.sh`; `python3 devBenches/base-image/update-upstream.py check` exits 0 on this branch; `git diff --stat origin/main...HEAD` still touches exactly the same 5 files. `shellcheck` is not on this machine's PATH.

## CI-driven follow-up (`fef6916`)

The first push of the fix round (`a4bdbe8`) failed CI's `regression` job: scenario (k)'s setup ran `update-upstream.py apply --source openrepotools --remove resume --yes`, unguarded, inside this `set -euo pipefail` test. `apply` calls `gh api` to prove the target commit is reachable from openRepoTools' default branch; that CI job has no `GH_TOKEN`, so `gh api` failed, `apply` exited 2 on its own Refusal, and the unguarded call took the whole test down with it (exit 2, no PASS/FAIL for the scenario). The reviewer's own instruction for (k) was already "edit the copied pin file by hand to drop the three `resume` lines, delete the copied file" — switched to exactly that (a small python3 edit + `rm`), which is both truer to the instruction and removes the network dependency entirely, rather than adding an unguarded network call the rest of this offline suite doesn't have. 76/76 green after, confirmed on CI (`fef6916`).

## Eagle pre-check (unchanged from the original PR; not re-run this round)

`openRepoShape` and `resume` identical to vendored; `openRepoTools` and `park` differ (a separate, already-merged PR bumped the pin since — see #39/#38). The real installer has still not been run on this branch; that is left for the acceptance pass after review, as before.

### DIFF
--- a/.github/workflows/speckit-git-bash.yml
+++ b/.github/workflows/speckit-git-bash.yml
@@ -17,6 +17,9 @@ on:
       - devBenches/base-image/update-upstream.py
       - devBenches/base-image/files/openreposhape/**
       - devBenches/base-image/files/openrepotools/**
+      - scripts/setup-estate-commands.sh
+      - devcontainer.test/test-setup-estate-commands.sh
+      - setup.sh
   pull_request:
     paths:
       - .github/workflows/speckit-git-bash.yml
@@ -33,6 +36,9 @@ on:
       - devBenches/base-image/update-upstream.py
       - devBenches/base-image/files/openreposhape/**
       - devBenches/base-image/files/openrepotools/**
+      - scripts/setup-estate-commands.sh
+      - devcontainer.test/test-setup-estate-commands.sh
+      - setup.sh
   workflow_dispatch:
 
 permissions:
@@ -66,6 +72,9 @@ jobs:
 
       - name: Check the vendored upstream copies against their pin
         run: python3 ./devBenches/base-image/update-upstream.py check
+
+      - name: Run the setup-estate-commands regression test
+        run: bash ./devcontainer.test/test-setup-estate-commands.sh
 
       - name: Run bootstrap suite
         run: bash ./devBenches/devcontainer.test/test-openspeckit-bootstrap.sh
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ setup.sh
   └── Summary + log file path
 ```
 
-Your host's `openRepoShape`, `openRepoTools`, `park` and `resume` commands come from workBenches' own pin (`devBenches/base-image/upstream-pin.yaml`). `setup.sh` places them from the vendored copies in `devBenches/base-image/files/openreposhape/` and `devBenches/base-image/files/openrepotools/`, and re-places them whenever a host copy differs from the pin, in either direction — behind, ahead, or hand-edited. To move them to a new upstream commit, move the pin with `update-upstream.py apply` and re-run `setup.sh`; do not edit the vendored files or the installed commands directly. Set `WORKBENCHES_SKIP_ESTATE_COMMANDS=1` to skip this step.
+Your host's `openRepoShape`, `openRepoTools`, `park` and `resume` commands come from workBenches' own pin (`devBenches/base-image/upstream-pin.yaml`). `setup.sh` places them from the vendored copies in `devBenches/base-image/files/openreposhape/` and `devBenches/base-image/files/openrepotools/`, and re-places them whenever a host copy differs from the pin, in either direction — behind, ahead, or hand-edited. To move them to a new upstream commit, move the pin with `update-upstream.py apply` and re-run `setup.sh`; do not edit the vendored files or the installed commands directly. The step refuses to run over a symlinked, non-regular, or read-only target, and verifies the four placed files against the vendored copies before it reports success. Set `WORKBENCHES_SKIP_ESTATE_COMMANDS=1` to skip this step.
 
 ## Wave Terminal Widgets
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ After setup, open any bench in VS Code → "Reopen in Container" to start develo
 
 ### Re-running setup.sh
 
-Safe to run repeatedly. Installed benches show `✓ up to date` and are skipped. Only new selections or missing images trigger builds.
+Safe to run repeatedly. Installed benches show `✓ up to date` and are skipped. Only new selections or missing images trigger builds. The estate commands (`openRepoShape`, `openRepoTools`, `park`, `resume`) are checked against workBenches' pin on every run and are re-placed only when a host copy differs from it.
 
 ## Docker Image Layers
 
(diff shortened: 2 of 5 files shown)


brettheap:
LANDED — lane xfactory-2 (xFactory-2), session de233a3a@Eagle, 2026-09-10T22:04Z, PR #40 → 83c3fed1a042efab1b011183593f727fe120534c into opensoft/workBenches main (merge commit, the repository's convention).

Review record: Opus adversarial review at 02bad7c found six reproduced defects (symlinked target written through; directory target copied into and reported placed; unpinned network fetch reachable when a pin row is dropped; partial sets on read-only or dangling targets; check exit 2 conflated with drift; PATH warning triplicated) — all fixed in a4bdbe8 + fef6916 (pre-flight refusals, pin-row check for all four files, no-fetch sentinel on the shims' REPO/REF variables, post-install verification, exit-2 distinction; 76 assertions). Re-verification at fef6916: MERGE.

Acceptance on Eagle (`scripts/setup-estate-commands.sh` from main; `setup.sh` parses): `update-upstream.py check` 5 rows ok; openRepoShape pinned at 0fa73b7, openRepoTools pinned at ff1f01a (moved by #39 before this landed); four commands `already installed (unchanged)`; `cmp` silent ×4; `park --help` and `openRepoShape --help` run; exit 0.

Known residue, non-blocking, unclaimed: (i) the two installer-failure branches do not say the four commands may be in a mixed state (one line each); (ii) `file_mode_octal` aborts silently under `set -e` on a host with no `stat`; (iii) the exact-0755 post-check would refuse a correct install on a filesystem that cannot hold that mode (DrvFs-style). Raven's operator steps are now: clone workBenches, run `./setup.sh`, rebuild the image.

PR #44, opened 2026-09-11
claude-profile: an opt-in --lane hands the final Claude launch to lane-start

brettheap:
Lane: openreposhape-2 (openRepoShape-2)
Claim: https://github.com/opensoft/workBenches/issues/43#issuecomment-5628726609

Closes #43.

## Summary

- `base-image/files/claude-profile` gains an opt-in `--lane <repo>-<n>` (and
  `CLAUDE_LANE` fallback): when named, the `run` action's final
  `exec "$claude_bin" …` becomes `exec lane-start "$lane" -- <the same claude
  args>`, with `CLAUDE_BIN`/`CLAUDE_CONFIG_DIR` exported so `lane-start`
  starts the same Claude under the same profile.
- With no `--lane`/`CLAUDE_LANE`, behaviour is unchanged — verified
  byte-for-byte (Claude's argv and every tmux call identical to the
  pre-change script under matched inputs).
- A named lane whose `lane-start` is not on `PATH` refuses (exit 2, names the
  `opensoft/brett-wip` clone + `scripts/link-estates` fix) before
  `claude_run_is_interactive`/`start_profile_tmux` ever run, so no tmux
  session is created first.
- The interactive auto-tmux path re-invokes the launcher inside a new
  detached tmux session; `CLAUDE_LANE` is threaded through that
  re-invocation explicitly (not relied on via tmux's environment
  inheritance, which does not reliably hand a freshly-exported variable to a
  new session on an already-running server).
- `--lane`/`-h`/`--help` are recognised only as leading options, before the
  action/profile, so anything after the profile name still reaches Claude
  untouched (e.g. `claude-profile work --help` still shows Claude's help).

## Test plan

- [x] New `devcontainer.test/test-claude-profile-lane-start.sh` (fake tmux,
      `lane-start`, `claude` on `PATH`): `--help` documents `--lane`/
      `CLAUDE_LANE`; no lane leaves `lane-start` uninvoked and Claude's args
      unchanged; `--lane` hands off with `CLAUDE_BIN`/`CLAUDE_CONFIG_DIR`
      set; a missing `lane-start` refuses before any tmux command runs;
      `CLAUDE_LANE` (env, no flag) behaves the same as `--lane`.
- [x] Existing launcher regression tests re-run against the patched script:
      `test-claude-tmux-statusline.sh`, `test-claude-profile-preserves-model.sh`,
      `test-shared-mcp-profiles.sh` — all pass.
[...]

PR #53, opened 2026-09-11
devcontainer.test/test-claude-tmux-statusline.sh: make the fake tmux/claude sandbox authoritative over the caller's real tmux identity

### LINKED ISSUE #52: test-claude-tmux-statusline.sh leaks the caller's real tmux identity into its fake sandbox, silently dying when run inside a live tmux session
Claimed by this lane (session 8706b0f5), branch fix/statusline-test-tmux-env-isolation.

## Reproduction

This lane runs inside a live tmux-backed Claude Code session, so its shell already carries real tmux identity: `TMUX=/tmp/tmux-1000/default,97960,390`, `TMUX_PANE=%390`, `WORKBENCHES_CLAUDE_TMUX_CHILD=1`, `WORKBENCHES_TMUX_SESSION=claude-team-05d-20260911090104-3553862`, `WORKBENCHES_TMUX_PANE=%390`.

**1) As-is** — `bash devcontainer.test/test-claude-tmux-statusline.sh` — exits 1 with **no output at all**, not even a `FAIL: ...` message. `bash -x` shows the script dies immediately after the first assertion's setup line — the `script -qefc "$tty_command" /dev/null` call that drives the launcher under a fake tty — and never reaches any `grep ... || fail "..."` line.

Root cause, isolated with a standalone repro (same fake `claude`/`tmux` on `PATH` as the test's own sandbox, launcher invoked directly under `script` with the real `TMUX` inherited): the *launcher itself* (`claude-profile`) aborts. `claude_run_is_interactive()` reads `TMUX`/`WORKBENCHES_CLAUDE_TMUX_CHILD` straight from the environment (`base-image/files/claude-profile:137-139`); with both inherited from the real caller, it returns 1 (non-interactive), so `start_profile_tmux` is skipped — but `export_tmux_identity()` (`claude-profile:180-188`) *also* reads `TMUX` directly, sees it set (from the real caller), and unconditionally shells out to whatever `tmux` is on `PATH` to resolve `#S` / `#{pane_id}`. In the test's sandbox that's the fake `tmux` — a dumb logger that only appends its argv to a log file and prints nothing to stdout — so `session`/`pane` both come back empty, and the function's last statement (`[[ -n "$pane" ]] && export WORKBENCHES_TMUX_PANE=...`) evaluates false. Under the launcher's own `set -euo pipefail`, that kills the launcher before it ever execs the fake `claude` (confirmed: the fake claude's log file is never created). `script -e` propagates that non-zero exit, and the *test's* own `set -euo pipefail` aborts the whole test script right there, silently, before any assertion runs.

**2) With the caller's tmux identity removed** — `env -u TMUX -u TMUX_PANE -u WORKBENCHES_CLAUDE_TMUX_CHILD -u WORKBENCHES_TMUX_SESSION -u WORKBENCHES_TMUX_PANE bash devcontainer.test/test-claude-tmux-statusline.sh` — passes: `claude tmux statusline tests passed`, exit 0.

So the test only works when launched from a shell with no tmux identity of its own — e.g. the fresh container shell `devcontainer.test/test.sh` runs it from — and silently dies (no assertion message, just exit 1) whenever it's run from inside any real tmux session, including the live Claude Code tmux sessions this estate's lanes run in every day.

## Not the #48 leak

PR #48 fixed a *different* leak in this same test file: `$HOME`-scoped usage-snapshot artifacts written by the statusline renderer during the test leaking into the caller's real `$HOME/.claude/usage-snapshots`. That fix pins `HOME`/`CLAUDE_CONFIG_DIR` to throwaway directories for the statusline-renderer invocations (this file's status-panel checks, roughly lines 87-110) and is unrelated to tmux identity.

This issue is about the *launcher*-invoking assertions earlier in the same file: the interactive tmux-session-creation check (~lines 61-77) and the non-interactive `mcp list` check (~lines 79-85). Both invoke `base-image/files/claude-profile` through the test's `common_env` array, which never touches `TMUX` / `WORKBENCHES_CLAUDE_TMUX_CHILD` / `WORKBENCHES_TMUX_SESSION` / `WORKBENCHES_TMUX_PANE`, so whatever the caller's shell already has leaks straight through.

Notably, PR #44 (which added the sibling test `devcontainer.test/test-claude-profile-lane-start.sh`) already engineered exactly this protection for *that* new file:

    unset TMUX WORKBENCHES_CLAUDE_TMUX_CHILD WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE TMUX_PANE 2>/dev/null || true

with a comment calling out this exact hazard ("a shell that is itself a claude-profile-launched tmux child ... or already inside some other tmux session") — but didn't retrofit the older `test-claude-tmux-statusline.sh` with the same guard. Noted in passing, left unfixed.

## Fix

Make the test's own sandbox authoritative over whatever tmux identity the calling shell happens to have, by unsetting every tmux-related variable the launcher and statusline consult before the sandbox runs — mirroring the guard already in `test-claude-profile-lane-start.sh`, extended to also cover `WORKBENCHES_CLAUDE_TMUX` (the opt-out switch `claude_run_is_interactive` also reads) and `CLAUDE_LANE` (the launcher's separate lane-start hand-off switch — see Update below):

    unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
        WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE CLAUDE_LANE

placed near the top of `test-claude-tmux-statusline.sh`, before the sandbox is built. The test's existing per-call `env -u TMUX -u TMUX_PANE -u WORKBENCHES_TMUX_SESSION -u WORKBENCHES_TMUX_PANE` for the "non-tmux panel" statusline assertion (current lines ~104-107) stays as-is (redundant with the top-level unset now, but harmless, and keeps that assertion self-documenting on its own).

Will re-run every `devcontainer.test/test-claude-*.sh` suite both from inside this tmux session and with the same variables unset, to confirm nothing else regresses, and check `.github/workflows/` + `devcontainer.test/test.sh` to confirm this doesn't change how CI/the image build invokes the test.

## Update (Copilot finding on PR #53, 2026-09-11T13:19:55Z, commit 4647cab)

The fix as originally landed missed one more input the launcher reads:
`CLAUDE_LANE` (`base-image/files/claude-profile:260`, `lane="${CLAUDE_LANE:-}"`),
read at the launcher's own top level, before any tmux/interactive check. If
a caller's shell has `CLAUDE_LANE` set (this lane's own session included —
`CLAUDE_LANE` is exactly the shape of variable a lane session plausibly
exports), the launcher's `run` action checks `command -v lane-start` before
`claude_run_is_interactive`/`start_profile_tmux` ever run:

- `lane-start` not on `PATH`: refuses with exit 2 and a stderr message the
  test's own `script -qefc ... /dev/null >/dev/null` wrapper swallows — the
  same silent-abort shape as the original bug, just a different trigger.
- `lane-start` on `PATH` (confirmed true in this lane's own session —
  `command -v lane-start` resolves to `/home/brett/.local/bin/lane-start`):
  the launcher hands the *entire* launch to it
  (`claude-profile:379-384`, `exec lane-start "$lane" -- ...`) instead of
  the fake `claude`. A test run could invoke real lane tooling.

Reproduced safely (without ever letting a real `lane-start` run) by checking
out the pre-fix-round test content and running it with
`CLAUDE_LANE=openRepoShape-2` exported and `PATH` constrained to exclude the
real `lane-start` — exit 2, zero output, same silent-abort-before-any-assertion
shape as the original finding.

Landed as a second commit on the same branch/PR: `CLAUDE_LANE` added to the
same `unset` block (now `unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX
WORKBENCHES_CLAUDE_TMUX_CHILD WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE
CLAUDE_LANE`). Verified with `CLAUDE_LANE=openRepoShape-2` exported both
inside this lane's live tmux session and with the tmux variables also
unset, plus a full re-run of the other `test-claude-*.sh` suites — all
green.

brettheap:

Closes #52.

## Problem

`devcontainer.test/test-claude-tmux-statusline.sh` builds a fake
`claude`/`tmux` sandbox and drives `base-image/files/claude-profile` and
`claude-statusline-command.sh` against it, but its launcher-invoking
assertions (the interactive tmux-session-creation check and the
non-interactive `mcp list` check) never touched `TMUX`, `TMUX_PANE`,
`WORKBENCHES_CLAUDE_TMUX`, `WORKBENCHES_CLAUDE_TMUX_CHILD`,
`WORKBENCHES_TMUX_SESSION`, or `WORKBENCHES_TMUX_PANE` — every one of them a
variable `claude-profile` or `claude-statusline-command.sh` reads straight
from the environment. Run the test from inside any real tmux session (a
live `claude-profile`-launched Claude Code pane included) and the caller's
own values for those variables leak straight through.

That's more than a mismatched assertion: with `TMUX` inherited but only the
fake `tmux` binary on `PATH`, the launcher's `claude_run_is_interactive()`
sees `TMUX` set and treats the launch as non-interactive, while
`export_tmux_identity()` *also* sees `TMUX` set and unconditionally shells
out to `tmux` to resolve `#S`/`#{pane_id}`. The fake `tmux` only logs its
argv and prints nothing, so both come back empty, and
`export_tmux_identity()`'s last statement —
`[[ -n "$pane" ]] && export WORKBENCHES_TMUX_PANE=...` — evaluates false.
Under the launcher's own `set -euo pipefail` that kills the launcher before
it ever execs the fake `claude` binary; `script -e` propagates the failure,
and the test's own `set -e` aborts the whole file right there, silently,
before any `grep ... || fail` line runs.

Reproduced from inside this lane's own live tmux-backed session:
- **As-is:** `bash devcontainer.test/test-claude-tmux-statusline.sh` → exit
  1, zero output, not even a `FAIL: ...` message. `bash -x` shows it dying
  at the `script -qefc "$tty_command" /dev/null` line — the launcher itself
  aborts before the test ever reaches an assertion.
- **With the caller's tmux identity removed:**
  `env -u TMUX -u TMUX_PANE -u WORKBENCHES_CLAUDE_TMUX_CHILD -u
  WORKBENCHES_TMUX_SESSION -u WORKBENCHES_TMUX_PANE bash
  devcontainer.test/test-claude-tmux-statusline.sh` → passes:
  `claude tmux statusline tests passed`, exit 0.

This is a different bug from the one #48 fixed: #48 stopped the test's
*statusline* invocations from leaking `$HOME`-scoped snapshot artifacts into
the caller's real home directory. This is about the *launcher* invocations
earlier in the same file leaking the caller's tmux identity inward. PR #44
(which added the sibling `test-claude-profile-lane-start.sh`) already
engineered exactly this guard for that new file, with a comment calling out
the same hazard, but never retrofitted the older
`test-claude-tmux-statusline.sh` — noted in passing, left unfixed. See #52
for the full trace.

## Fix

Clear `TMUX`, `TMUX_PANE`, `WORKBENCHES_CLAUDE_TMUX`,
`WORKBENCHES_CLAUDE_TMUX_CHILD`, `WORKBENCHES_TMUX_SESSION`, and
`WORKBENCHES_TMUX_PANE` near the top of the test, before the sandbox is
built — mirroring the guard `test-claude-profile-lane-start.sh` already
carries, extended to also cover `WORKBENCHES_CLAUDE_TMUX` (the opt-out
switch `claude_run_is_interactive` also reads). The test's existing
per-call `env -u TMUX -u TMUX_PANE -u WORKBENCHES_TMUX_SESSION -u
WORKBENCHES_TMUX_PANE` around the non-tmux statusline panel assertion is
untouched (now redundant, still harmless, keeps that one assertion
self-documenting). Purely additive: 19 lines added, 0 removed, 0 changed.

## Test plan

- [x] `bash devcontainer.test/test-claude-tmux-statusline.sh` from inside
      this lane's own live tmux-backed Claude Code session — passes
      (`claude tmux statusline tests passed`, exit 0); repeated 3x, no
      flakiness.
- [x] `env -u TMUX -u TMUX_PANE -u WORKBENCHES_CLAUDE_TMUX_CHILD -u
      WORKBENCHES_TMUX_SESSION -u WORKBENCHES_TMUX_PANE bash
      devcontainer.test/test-claude-tmux-statusline.sh` — passes, same
      result.
- [x] Same test invoked with explicit launcher/statusline paths (the form
      `devcontainer.test/test.sh` uses inside the built image) — passes.
- [x] `devcontainer.test/test-claude-profile-lane-start.sh` — passes both
      inside this live tmux session and with the variables unset.
- [x] `devcontainer.test/test-claude-profile-preserves-model.sh` — passes
      both ways (unaffected either way: it only exercises the launcher's
      `status` action, which never reaches the tmux-sensitive code paths).
- [x] `devcontainer.test/test-claude-statusline-snapshots.sh` — passes both
      ways.
- [x] Confirmed `.github/workflows/speckit-git-bash.yml` and
      `speckit-git-powershell.yml` don't invoke any `test-claude-*.sh` suite
      (they only run the speckit/git-bootstrap tests), so CI's own
      invocation of these tests is unaffected by this change.

## Summary by Sourcery

Isolate the Claude tmux statusline test from the caller's tmux and lane environment so it consistently exercises only the fake sandbox.

Bug Fixes:
- Prevent tmux and lane-related environment variables from leaking into the Claude tmux statusline test and causing false failures or unintended real-tool invocations.

Tests:
- Make the sandbox-based tmux statusline test authoritative and reliable when run from within an existing tmux or Claude lane session.

### DIFF
--- a/devcontainer.test/test-claude-tmux-statusline.sh
+++ b/devcontainer.test/test-claude-tmux-statusline.sh
@@ -13,21 +13,33 @@ trap 'rm -rf "$TEST_ROOT"' EXIT
 # This test drives the real launcher and statusline scripts against a fake
 # tmux/claude sandbox built below, then asserts on the fake tmux/claude logs
 # that sandbox produces. Both scripts also read tmux identity straight from
-# the environment (WORKBENCHES_CLAUDE_TMUX*, TMUX*, WORKBENCHES_TMUX_*). If
-# this test itself is run from inside a real tmux session — a live
-# claude-profile-launched Claude Code pane included — those variables leak
-# in from the caller and are indistinguishable from the sandbox's own state.
-# That alone corrupts the assertions below, and it can also abort the
-# launcher outright: with a real TMUX inherited but only the fake tmux
-# binary on PATH, the launcher's export_tmux_identity() queries tmux, gets
-# back nothing the fake understands, and its last command's failure under
-# `set -e` kills the launcher before it ever reaches Claude. Mirrors the
-# same guard in test-claude-profile-lane-start.sh, extended to also cover
+# the environment (WORKBENCHES_CLAUDE_TMUX*, TMUX*, WORKBENCHES_TMUX_*), and
+# the launcher separately reads CLAUDE_LANE to decide whether to hand the
+# launch off to `lane-start` instead of the fake claude. If this test itself
+# is run from inside a real tmux session — a live claude-profile-launched
+# Claude Code pane included — those variables leak in from the caller and
+# are indistinguishable from the sandbox's own state. That alone corrupts
+# the assertions below, and it can also abort the launcher outright: with a
+# real TMUX inherited but only the fake tmux binary on PATH, the launcher's
+# export_tmux_identity() queries tmux, gets back nothing the fake
+# understands, and its last command's failure under `set -e` kills the
+# launcher before it ever reaches Claude. A leaked CLAUDE_LANE is worse in a
+# different way: if a real `lane-start` happens to be on PATH (as it is in
+# exactly the lane sessions this test is likeliest to be run from), the
+# launcher hands the whole launch to it instead of the fake claude — a test
+# run invoking real lane tooling. Mirrors the same guard in
+# test-claude-profile-lane-start.sh, extended to also cover
 # WORKBENCHES_CLAUDE_TMUX (the opt-out switch claude_run_is_interactive
-# reads). Clear all of it before the sandbox runs, so the sandbox's own
-# state is always authoritative.
+# reads) and CLAUDE_LANE (the launcher's lane-start hand-off switch). Clear
+# all of it before the sandbox runs, so the sandbox's own state is always
+# authoritative. (CLAUDE_BIN/CLAUDE_PROFILES_HOME/CLAUDE_PROFILES_MANIFEST/
+# WORKBENCHES_SHARED_MCP_FAMILIES are already owned below via common_env's
+# explicit assignments on every launcher invocation, so an inherited value
+# for any of those is always overridden rather than leaked; CLAUDE_CONFIG_DIR
+# is only ever written by the launcher for its children, never read as one
+# of its own inputs, so nothing to clear there either.)
 unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
-    WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE 2>/dev/null || true
+    WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE CLAUDE_LANE 2>/dev/null || true
 
 fail() {
     echo "FAIL: $*" >&2
--- a/devcontainer.test/test-claude-tmux-statusline.sh
+++ b/devcontainer.test/test-claude-tmux-statusline.sh
@@ -10,6 +10,25 @@ STATUSLINE="${2:-$REPO_ROOT/base-image/files/claude-statusline-command.sh}"
 TEST_ROOT="$(mktemp -d)"
 trap 'rm -rf "$TEST_ROOT"' EXIT
 
+# This test drives the real launcher and statusline scripts against a fake
+# tmux/claude sandbox built below, then asserts on the fake tmux/claude logs
+# that sandbox produces. Both scripts also read tmux identity straight from
+# the environment (WORKBENCHES_CLAUDE_TMUX*, TMUX*, WORKBENCHES_TMUX_*). If
+# this test itself is run from inside a real tmux session — a live
+# claude-profile-launched Claude Code pane included — those variables leak
+# in from the caller and are indistinguishable from the sandbox's own state.
+# That alone corrupts the assertions below, and it can also abort the
+# launcher outright: with a real TMUX inherited but only the fake tmux
+# binary on PATH, the launcher's export_tmux_identity() queries tmux, gets
+# back nothing the fake understands, and its last command's failure under
+# `set -e` kills the launcher before it ever reaches Claude. Mirrors the
+# same guard in test-claude-profile-lane-start.sh, extended to also cover
+# WORKBENCHES_CLAUDE_TMUX (the opt-out switch claude_run_is_interactive
+# reads). Clear all of it before the sandbox runs, so the sandbox's own
+# state is always authoritative.
+unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
+    WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE 2>/dev/null || true
+
 fail() {
     echo "FAIL: $*" >&2
     exit 1


brettheap:
@codex review

brettheap:
@copilot-pull-request-reviewer thanks — this was a real gap, not a false positive.

`CLAUDE_LANE` is a separate input `claude-profile` reads at its own top level (`lane="${CLAUDE_LANE:-}"`, `claude-profile:260`), before any tmux/interactive check runs. The original commit's unset block only covered the tmux-related variables and missed it. Confirmed both failure modes it enables:

- if `lane-start` isn't on `PATH`: the launcher refuses with exit 2 and a stderr message the test's own `script -qefc ... /dev/null >/dev/null` wrapper swallows — the same silent-abort-before-any-assertion shape as the original bug.
- if `lane-start` *is* on `PATH` (true in this lane's own session — `command -v lane-start` resolves to `/home/brett/.local/bin/lane-start`): the launcher hands the whole launch to it (`claude-profile:379-384`) instead of the fake claude.

Reproduced the first case safely (constrained `PATH` so the real `lane-start` was never reachable, so nothing real ever ran) rather than the second, since triggering the second for real is exactly the unsafe behavior being fixed.

Fixed in 35a804a: added `CLAUDE_LANE` to the same `unset` block, and documented in that commit (and in the block's own comment) why the launcher's other lane/launch-shaping inputs don't need the same treatment — `CLAUDE_BIN`, `CLAUDE_PROFILES_HOME`, `CLAUDE_PROFILES_MANIFEST`, and `WORKBENCHES_SHARED_MCP_FAMILIES` are already owned via `common_env`'s explicit assignments on every launcher invocation, and `CLAUDE_CONFIG_DIR` is only ever written by the launcher for its children, never read as one of its own inputs. Verified `test-claude-tmux-statusline.sh` passes with `CLAUDE_LANE=openRepoShape-2` exported, both inside this lane's live tmux session and with the tmux variables also unset, plus a clean re-run of the other `test-claude-*.sh` suites.

brettheap:
LANDING — lane openreposhape-2 (openRepoShape-2), session 8706b0f5, 2026-09-11T13:31:26Z, PR #53 into opensoft/workBenches main (test-claude-tmux-statusline.sh isolates the launcher from the caller's tmux identity; closes workBenches#52)

PR #51, opened 2026-09-11
devBenches/base-image: re-vendor openRepoTools at f519f31 — `status`, the fourth command, joins park and resume; bare `park` asks before sweeping, `park --all`

brettheap:
The `openrepotools` source pin moves from ff1f01aae70b8f21dce0e4654da071a1281b41a4
(pinned in #39) to f519f31c28fad3ab2f54cd60d828226d866db3f5 —
`gh api repos/opensoft/openRepoTools/compare/main...f519f31c28fad3ab2f54cd60d828226d866db3f5`
says `identical`, so it is the commit opensoft/openRepoTools's main carries after
opensoft/openRepoTools#15.

Between the two commits (opensoft/openRepoTools#8-#15, all squash-merged):
`status` is the FOURTH command, read-only, runs nothing, in five layers — the
local layer (ahead/behind as of the last fetch, dirty, branches, worktrees, legs
against their pins), `--fetch` (the one write bounded to remote-tracking refs by
a pinned refspec), a fork against a remote named `upstream` or, under `--fetch`,
one read-only `gh api` question, the shape pin (contracts/shape-pin.yaml
re-hashed, the pin read against the standard's main) and the parked record
against disk (#9-#13, Brett Heap's rulings of 2026-09-10 and 2026-09-11); bare
`park` parks the estate around the cwd and ASKS before sweeping, `park --all` /
`-a` is the sweep unasked (#8, his ruling of 2026-09-10); `park` and `resume`
refuse a value-less flag with exit 2 in their own words (#14); `park` and
`status` read `.gitmodules` with `git config -z` (#15). `openRepoTools --install`
now places four files, so its vendored shim changes too.

So this is one `apply --add status` run AND the consumers that enumerate the
commands, each gaining the fifth name: the Dockerfile COPYs and chmods
/usr/local/bin/status beside the other three; scripts/setup-estate-commands.sh
pre-flights a fifth target, requires its pin row and post-verifies it;
setup.sh's and README's lists name it; and
devcontainer.test/test-setup-estate-commands.sh asserts the fifth placement,
its second-run `already installed` line, its byte-identity to the vendored copy
and its 0755 mode. update-upstream.py is untouched.

check: exit 0, 6 rows across 2 sources.
devcontainer.test/test-setup-estate-commands.sh: 83 PASS (was 76), GREEN, exit 0.
devBenches/devcontainer.test/test-speckit-git-feature.sh: 56 PASS, GREEN, exit 0.
[...]

PR #18, opened 2026-09-01
Advance merged pyBench and cloudBench submodules

brettheap:
## Summary
- advance pyBench to merged PR #3 commit 59ad3a7c
- advance cloudBench to merged PR #4 commit d91e08fe
- leave rustBench unchanged because main already points at b34fab7a

## Verification
- both gitlinks equal the child repositories current main SHA
- child PRs were merged from exact tested heads
- git diff --check passes

Sonar is advisory for this cleanup because the organization is over its licensed LOC allocation.

## Summary by Sourcery

Enhancements:
- Advance the pyBench and cloudBench submodules to their current merged main revisions while leaving rustBench unchanged.

PR #17, opened 2026-09-01
Preserve and separate workBenches follow-up changes

brettheap:
## Summary

This stacked cleanup preserves and separates the previously uncommitted workBenches follow-ups:

- ignore machine-local CLI/runtime state
- harden Claude and Codex profile runtimes
- add OpenCode and shared MCP profiles
- add protected AI credential backup tooling
- add the shared Vibe Annotations service
- integrate safe bench startup without regressing live-container or rustBench safety
- add the Windows audio session watcher
- capture the workBenches-next architecture and pending OpenCode OpenSpec changes

## Stack

- base: `main` (workBenches PRs #3, #15, and #16 are merged)
- child subrepo heads are not changed by this PR; the pyBench dependency was merged through workBenches PR #16

## Validation

- Bash syntax on all 20 changed shell scripts
- ShellCheck with documented exclusions for pre-existing or intentional findings
- JSON validation for the credential manifest example
- Vibe Annotations Compose rendering
- Claude tmux/statusline regression tests
- Codex profile runtime override tests
- shared MCP profile tests
- Vibe Annotations shared-service tests
- Wave launcher lifecycle and start-during-check race tests
- OpenSpec validation: 4 passed, 0 failed
- PowerShell AST parse for the audio watcher
- credential backup `--help` smoke test

SonarQube analysis is advisory because the current licensed LOC ceiling may prevent it from running. No live bench was recreated, and no Azure or Microsoft 365 credential/tenant mutation was performed.

PR #16, opened 2026-09-01
Make pyBench SonarQube MCP routing durable

brettheap:
## Summary

- make pyBench SonarQube MCP routing durable in the Codex profile launcher
- add the associated runtime-routing specification and implementation artifacts
- add a targeted regression test for the runtime override

## Stack

- base: `main` (workBenches PR #15 is merged)
- dependencies: workBenches PRs #3 and #15 are merged

## Validation

- `bash -n base-image/files/codex-profile devcontainer.test/test-codex-profile.sh`
- `shellcheck base-image/files/codex-profile devcontainer.test/test-codex-profile.sh`
- `devcontainer.test/test-codex-profile.sh`

SonarQube is advisory because the current licensed LOC ceiling may prevent analysis.

## Summary by Sourcery

Make pyBench SonarQube MCP access durable across launcher use, container restarts, and recreation while retaining the host profile and private proxy boundary.

New Features:
- Add durable runtime-specific SonarQube MCP routing for pyBench through the Codex profile launcher and private shared network configuration.

Bug Fixes:
- Allow Codex running in pyBench to reach the private SonarQube MCP proxy without modifying the shared host profile.

Enhancements:
- Preserve existing authentication, profile selection, user arguments, unrelated MCP settings, and proxy security boundaries while applying one-process URL overrides.

Documentation:
- Add the feature specification, runtime-routing contract, implementation plan, data model, research, validation quickstart, and task tracking artifacts.

Tests:
- Add focused regression coverage for runtime override presence, absence, quoting, argument preservation, and credential non-forwarding.
- Integrate the Codex profile runtime-routing test into the devcontainer test suite.

Chores:
- Mark the OpenSpec-to-Speckit implementation handoff complete.

PR #54, opened 2026-09-11
Re-vendor openRepoTools at a2c8521: `status`'s record layer reads a deleted worktree as none, one workspace as one record, never sends a `--no-push` record to `resume`; every value-taking flag refuses an empty value

brettheap:
The `openrepotools` source pin moves from f519f31c28fad3ab2f54cd60d828226d866db3f5
(pinned in #51) to a2c85219942eb8a8fbb2cfd20131e0ebf0336c10 —
`gh api repos/opensoft/openRepoTools/compare/main...a2c85219942eb8a8fbb2cfd20131e0ebf0336c10`
says `identical`, so it is the commit opensoft/openRepoTools's main carries after
opensoft/openRepoTools#18.

Between the two commits (opensoft/openRepoTools#16, #17 and #18, all squash-merged):
openRepoTools pins openRepoShape at 32450eb, its main of 2026-09-11, the commit this
repository's own `openreposhape` row already carries (#16; none of the eight mounted
files moved). Every value-taking flag of `park`, `resume` and `status` refuses an
EMPTY value in the same words as a missing one, exit 2, so `--lane ""` no longer parks
with the lane dropped and `--repo ""` no longer falls back to the walk-up (#17). And
`status`'s parked-record layer, the follow-up Copilot's two reviews of #13 were owed
(#18): a recorded feature whose worktree is gone is "no worktree here", not clean; a
registration `git worktree list` still holds for it is told apart from none, and the
finding names `git -C <repo> worktree prune` before `resume <Name>`, `git worktree
unlock <path>` first when the block is locked, since prune skips a locked one, and a
leftover directory is moved aside, never deleted; a block git itself calls `prunable`
is stale whatever is on disk, and a registered path is a worktree only while its own
`.git` is there — the record layer, the unrecorded sweep and the local layer's
`read_worktrees` now apply the one rule; two checkouts of one workspace repository
are one record, keyed on `<org>/<basename>` with both halves case-folded, in `status`
and `resume` alike; a `--no-push` record with no worktree here is one line naming the
only exit and never names `resume`; and `status --fetch` fetches every repository of
the estate FIRST, before the first row prints, keyed by physical path, so a parked
commit that lived only on a leg's origin is no longer reported missing and then
fetched further down the same report. The installer places the same four files.

So this is one `apply` run and nothing else: no command was added or removed, so
the Dockerfile COPYs, scripts/setup-estate-commands.sh, setup.sh, the README and the
test scripts that enumerate the commands are untouched. update-upstream.py is untouched.

devBenches/devcontainer.test/test-speckit-git-feature.sh: 56 PASS, GREEN, exit 0.
The base image was not rebuilt here; only the vendored copies and their pin rows change.

## Summary by Sourcery

Re-vendor openRepoTools to improve worktree status accuracy, workspace record consolidation, fetch behavior, and command-line validation.

Bug Fixes:
- Ensure status consistently distinguishes missing, stale, locked, and registered worktrees, including safe handling of leftover directories and proper resume guidance.
- Make status --fetch update all repositories before reporting results and consolidate multiple checkouts of the same workspace into one record.
- Reject empty values for all value-taking park, resume, and status flags.
- Prevent status from suggesting resume for worktree-less --no-push records.

Enhancements:
- Re-vendor openRepoTools and update its upstream commit and file checksums.

### DIFF
--- a/devBenches/base-image/files/openrepotools/park
+++ b/devBenches/base-image/files/openrepotools/park
@@ -889,16 +889,30 @@ while [ $# -gt 0 ]; do
     # anywhere, with no question (the RULING of 2026-09-10 at the sweep's
     # section head: "park all should be park -a or park --all").
     -a | --all) PARK_ALL=1; shift ;;
-    # A MISSING VALUE IS THIS COMMAND'S REFUSAL, exit 2, in its own words:
-    # `${2:?…}` would be bash's message and bash's exit 1, which is not a
-    # status this command means anything by (`status` gave the reason first:
+    # A MISSING OR EMPTY VALUE IS THIS COMMAND'S REFUSAL, exit 2, in its own
+    # words: `${2:?…}` would be bash's message and bash's exit 1, which is not
+    # a status this command means anything by (`status` gave the reason first:
     # its 1 is "findings"; here, a typo must not read like `make park`'s own
     # exit).
+    # EMPTY IS REFUSED THE SAME WAY, which is why every arm below asks for a
+    # NON-EMPTY `$2` and not only for a second argument. `${2:?…}` rejected
+    # unset AND null; `[ $# -ge 2 ]` alone rejects only the first, and what it
+    # lets through is read later as "was this given at all?" — so
+    # `park --lane ""` would carry on with `LANE` empty, the lane is then left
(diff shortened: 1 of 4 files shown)


brettheap:
@codex review

PR #56, opened 2026-09-11
devcontainer.test/test-claude-tmux-statusline.sh: subprocess regression for the env-isolation guard

### LINKED ISSUE #55: devcontainer.test/test-claude-tmux-statusline.sh: the env-isolation guard has no subprocess regression, so test.sh can't tell it apart from no guard at all

## Finding (Copilot, PR #53, `devcontainer.test/test-claude-tmux-statusline.sh:42`)

Copilot's final review on PR #53 (commit 35a804a, submitted
2026-09-11T13:31:30Z) left this suppressed comment on the guard's `unset`
line, verbatim:

> This guard is not exercised by the standard `devcontainer.test/test.sh`
> run: that entry point invokes this script in a clean container, and the
> script immediately unsets the variables before any assertion. A future
> removal or omission from this list would therefore still pass that test
> run; please add a wrapper/subprocess regression that invokes the test
> with representative tmux/lane variables set and fails on the pre-fix
> behavior, in addition to the manual live-tmux run.
>
> ```
> unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
>     WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE CLAUDE_LANE 2>/dev/null || true
> ```

## Why it was missed

PR #53 (closing #52) merged as 70abe6513552db164c295cf42981efb5f66e7cfd.
Copilot's review carrying this finding was submitted at
**2026-09-11T13:31:30Z**; the PR was merged at **2026-09-11T13:31:59Z** — 29
seconds later. There was no window between the review landing and the merge
for it to be read, let alone actioned, so the residual rode the merge
untouched.

The finding itself is correct and not a false positive: `test.sh` runs
`test-claude-tmux-statusline.sh` from a clean container shell that never
had `TMUX`/`WORKBENCHES_CLAUDE_TMUX*`/`WORKBENCHES_TMUX_*`/`CLAUDE_LANE` set
in the first place, so the guard's `unset` line is a no-op in that run —
removing it, or dropping one name from its list, would not turn that run
red. The only proof the guard does anything currently lives in PR #53's own
manual, non-repeatable verification (running the test inside a live tmux
session by hand).

## What will land

A subprocess regression that sets representative tmux/lane variables
(the shape of a live `claude-profile`-launched, lane-started Claude Code
pane: a real-looking `TMUX` socket/pane pair, both `WORKBENCHES_CLAUDE_TMUX`
switches, `WORKBENCHES_TMUX_SESSION`/`WORKBENCHES_TMUX_PANE`, and
`CLAUDE_LANE`) in the *calling* environment before invoking
`test-claude-tmux-statusline.sh`, and fails unless that test still passes —
so a future removal or omission from the guard's `unset` list turns this
regression red even under the plain `test.sh` container run. `CLAUDE_LANE`
is handled safely regardless of the guard's state: the regression pins
`PATH` to exclude any real `lane-start`, so a regressed guard can only reach
`claude-profile`'s existing "lane-start missing" refusal, never a real
hand-off.

The new regression will be verified to actually fail on the pre-fix
behaviour (temporarily removing the guard's `unset` line, confirming this
regression goes red, then restoring it and confirming green) before it
lands, per the finding's own request.

brettheap:
Claim: https://github.com/opensoft/workBenches/issues/55#issuecomment-5637339061

Closes #55.

## What was wrong

PR #53 (closing #52) added an environment-isolation guard to the top of
`devcontainer.test/test-claude-tmux-statusline.sh`:

```
unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
    WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE CLAUDE_LANE 2>/dev/null || true
```

Copilot's final review on that PR (commit 35a804a, submitted
**2026-09-11T13:31:30Z**) left this finding on that line, verbatim:

> This guard is not exercised by the standard `devcontainer.test/test.sh`
> run: that entry point invokes this script in a clean container, and the
> script immediately unsets the variables before any assertion. A future
> removal or omission from this list would therefore still pass that test
> run; please add a wrapper/subprocess regression that invokes the test
> with representative tmux/lane variables set and fails on the pre-fix
> behavior, in addition to the manual live-tmux run.

The PR merged at **2026-09-11T13:31:59Z** — 29 seconds after that review
landed — so it rode the merge untouched. #55 is the governing issue for
this residual.

## What lands

`devcontainer.test/test-claude-tmux-statusline-env-isolation.sh`, wired
into `test.sh` immediately after the existing "Claude profile tmux status
panel" line, the same way every other sibling test is wired in (the whole
`devcontainer.test/` directory is bind-mounted as `/test` in the built
image — no Dockerfile change needed). It sets representative values for
every name the guard unsets — a real-looking tmux socket/pane pair, both
`WORKBENCHES_CLAUDE_TMUX` switches, a lane session id, and a lane name,
mirroring the shape of a live `claude-profile`-launched, lane-started
Claude Code pane that #52 itself reproduced with — in the *calling*
environment, then runs `test-claude-tmux-statusline.sh` as a subprocess and
fails unless it still passes underneath that contamination.

`CLAUDE_LANE` is the one name on the list that is unsafe to leak with a
reachable real value: if the guard fails to unset it, `claude-profile`
hands the whole launch to a real `lane-start` before any tmux/interactive
check runs. This regression can't trigger that for real regardless of the
guard's state — its `PATH` is pinned to `/usr/bin:/bin`, which never
contains `lane-start` (a user-local install, e.g. `~/.local/bin`), so a
regressed guard can only ever reach `claude-profile`'s existing
"lane-start missing" refusal (exit 2) — the same safe pattern the
`refusal_env` scenario in `test-claude-profile-lane-start.sh` already uses.

## How it was proven

**Current proof, against b56c089's direct-probe assertion — this is the
primary evidence now, and it is what runs first:** for each of the seven
guarded names, removed exactly that one name from the inner test's guard,
ran the wrapper, and confirmed it failed naming that variable; restored the
guard (`diff` against a pristine backup came back empty) and confirmed the
wrapper passed again:

```
TMUX                          -> FAIL: guard did not unset: TMUX
TMUX_PANE                     -> FAIL: guard did not unset: TMUX_PANE
WORKBENCHES_CLAUDE_TMUX       -> FAIL: guard did not unset: WORKBENCHES_CLAUDE_TMUX
WORKBENCHES_CLAUDE_TMUX_CHILD -> FAIL: guard did not unset: WORKBENCHES_CLAUDE_TMUX_CHILD
WORKBENCHES_TMUX_SESSION      -> FAIL: guard did not unset: WORKBENCHES_TMUX_SESSION
WORKBENCHES_TMUX_PANE         -> FAIL: guard did not unset: WORKBENCHES_TMUX_PANE
CLAUDE_LANE                   -> FAIL: guard did not unset: CLAUDE_LANE
```

All seven restores confirmed byte-identical to the pre-edit backup. Also
re-ran the whole wrapper 3x against the intact guard (pass every time) and
the inner test once (pass) after the seven rounds.

**First round's evidence (005dc18, before the direct probe existed — kept
here as history, not as a description of today's behavior):** commenting
out the guard's *entire* `unset` statement and running the wrapper — which
at that commit had only the end-to-end assertion, nothing running ahead of
it — failed exactly as required:

```
FAIL: test-claude-tmux-statusline.sh did not survive representative tmux/lane contamination (exit 2):
```

— the wrapped subprocess exited **2** with **zero bytes** of output (the
inner test's `script -qefc "$tty_command" /dev/null >/dev/null` swallows
`claude-profile`'s stderr refusal, same silent-abort shape #52 itself
described). This end-to-end evidence still holds precisely under the
conditions it was taken: the whole guard removed, and nothing ahead of the
end-to-end assertion to intercept first. It is no longer what today's
(b56c089) wrapper actually surfaces, since its direct probe now runs first
and catches a missing guard before the end-to-end assertion is ever
reached — `could not locate the env-isolation guard's 'unset ...' line`
if the whole statement is gone, or that name's own `guard did not unset:
<NAME>` if only one name is missing.

Separately, in that same first round, running the *unmodified*
(guard-disabled) `test-claude-tmux-statusline.sh` directly from inside this
lane's own live tmux session — which already carries real
`TMUX`/`WORKBENCHES_CLAUDE_TMUX_CHILD`/`WORKBENCHES_TMUX_SESSION`/
`WORKBENCHES_TMUX_PANE` — reproduced the *original* #52 failure mode too:
exit 1, zero bytes of output. Restored the guard line (`diff` against the
pre-edit original came back identical) and reran: passed again, 3x with no
flakiness.

## Test plan

- [x] `bash devcontainer.test/test-claude-tmux-statusline.sh` — passes
      (`claude tmux statusline tests passed`, exit 0).
- [x] `bash devcontainer.test/test-claude-tmux-statusline-env-isolation.sh`
      (new) — passes (`claude tmux statusline env-isolation regression
      passed`, exit 0), repeated 3x, no flakiness.
- [x] Direct-probe proof (b56c089, current): each of the seven guarded
      names removed one at a time from the inner test's guard → wrapper
      fails naming that variable (`guard did not unset: <NAME>`) → guard
      restored (byte-identical) → wrapper passes again. All seven; see
      "How it was proven" above for the full list.
- [x] First round's evidence (005dc18, historical): whole guard
      temporarily disabled → wrapper (end-to-end assertion only, at that
      commit) failed (wrapped subprocess exit 2, zero output) → guard
      restored (byte-identical) → wrapper passed again.
- [x] `bash devcontainer.test/test-claude-profile-lane-start.sh` — passes.
- [x] `bash devcontainer.test/test-claude-profile-preserves-model.sh` —
      passes.
- [x] `bash devcontainer.test/test-claude-statusline-snapshots.sh` (on
      host) — passes.
- [x] Full `devcontainer.test/test.sh` built and run inside the real
      `workbench-base` container (`docker compose up` + `exec`): "Claude
      profile tmux status panel" and the new "Claude profile tmux
      statusline env-isolation regression" both pass. (One unrelated,
      pre-existing test in that same container run — "Claude statusline
      usage snapshots" — failed for reasons unconnected to this change,
      a usage-snapshot-file-path issue in `test-claude-statusline-snapshots.sh`,
      a file this PR never touches; left untouched, out of this PR's scope.)

## Summary by Sourcery

Strengthen Claude tmux statusline environment-isolation coverage against inherited tmux and lane state.

Enhancements:
- Add a regression test that verifies Claude tmux statusline tests remain isolated from inherited tmux and lane environment variables.
- Verify each environment variable in the isolation guard is actively cleared before running the statusline test.

Tests:
- Integrate the new environment-isolation regression into the devcontainer test suite.


brettheap:
@codex review

brettheap:

Addressed the finding from Copilot's review of 005dc18, in b56c089:

Copilot's finding: the wrapper's contamination did not actually make several
of the seven guarded names load-bearing. `claude-profile` never reads
`TMUX_PANE` or either `WORKBENCHES_TMUX_*` name at all; the inner test's own
two statusline assertions overwrite or explicitly unset both
`WORKBENCHES_TMUX_*` names regardless of the guard; and
`WORKBENCHES_CLAUDE_TMUX=1` doesn't change `claude_run_is_interactive`'s
answer (only `off` does). So the end-to-end subprocess run alone could still
pass with any of those four silently dropped from the guard's unset list.

**What changed:** added a direct assertion that runs *ahead of* the existing
end-to-end one. It locates the inner test's `unset ...` guard statement
(however many physical lines its backslash continuation spans; fails loudly
if it's no longer there), truncates a temp copy of the inner test's own
source right after it, appends a probe of all seven names
(`declare -p "$name"` under an `if`, not `&&`, so probing an already-unset
name can't itself trip `set -e`), and runs that probe under the same
contamination — asserting none of the seven names is still set in that
child's own environment. The probe runs first specifically so every one of
the seven mutation rounds below fails naming the missing variable, rather
than some of them being masked by the end-to-end check's more generic
symptom-only message (a handful of these names, if dropped, would also
break the end-to-end run, but with a message that doesn't say which name
caused it).

Building this also surfaced a real bug in the previous version of this same
file: `still_set="$(grep '^STILL-SET:' ... | cut ... | tr ...)"` aborted the
whole script silently whenever grep found zero matches — which is the
*common, expected* case (the guard doing its job). Under this script's own
`set -euo pipefail`, grep's "no match" exit 1 became the pipeline's exit
status, killing the script before it could report success. Same class of
bug as #52 itself, now inside the regression meant to catch it. Fixed with
`|| true` on that pipeline, and applied the same fix to the guard-line
lookup for the same reason.

**Proof, for each of the seven names:** removed exactly that one name from
the inner test's guard, ran the wrapper, confirmed it failed naming that
variable, restored the guard (diff against a pristine backup empty), and
confirmed the wrapper passed again:

```
TMUX                          -> FAIL: guard did not unset: TMUX
TMUX_PANE                     -> FAIL: guard did not unset: TMUX_PANE
WORKBENCHES_CLAUDE_TMUX       -> FAIL: guard did not unset: WORKBENCHES_CLAUDE_TMUX
WORKBENCHES_CLAUDE_TMUX_CHILD -> FAIL: guard did not unset: WORKBENCHES_CLAUDE_TMUX_CHILD
WORKBENCHES_TMUX_SESSION      -> FAIL: guard did not unset: WORKBENCHES_TMUX_SESSION
WORKBENCHES_TMUX_PANE         -> FAIL: guard did not unset: WORKBENCHES_TMUX_PANE
CLAUDE_LANE                   -> FAIL: guard did not unset: CLAUDE_LANE
```

All seven restores confirmed byte-identical to the pre-edit backup via
`diff`. Also re-ran the whole wrapper 3x against the intact guard (pass
every time) and the inner test once (pass), plus
`test-claude-profile-lane-start.sh` and
`test-claude-profile-preserves-model.sh` (no regressions).

Not merging; this thread is left for you to resolve or continue.

brettheap:

Updated the PR body's "How it was proven" section and one Test plan bullet
for Copilot's wording nit on b56c089 (no code change, head is still
b56c089). The record had gone stale: since the direct probe now runs
first, a missing guard entry fails at `guard did not unset: <NAME>` and
never reaches `test-claude-tmux-statusline.sh`, so describing the exit-2/
zero-output subprocess result as current proof was no longer accurate.

The section now leads with the current proof (the seven `FAIL: guard did
not unset: <NAME>` lines, one per guarded name, each with the guard
restored and re-passing), and keeps the original exit-2/zero-output
observation explicitly labelled as the first round's (005dc18) end-to-end
evidence — still true of the conditions it was taken under, but no longer
what today's wrapper surfaces first.

PR #20, opened 2026-09-01
Advance final merged workBench subrepos

brettheap:
## Summary

Advance the final merged subrepository heads:

- cloudBench PR #1: retire obsolete Amnezia VM tooling
- 365Bench PR #1: add cross-platform Microsoft business-app tooling with review and Sonar security fixes

## Exact child merges

- cloudBench: `955a8ece6ac773dedb63b93013a93200ef506aa3`
- 365Bench: `09ddb50b17228bc750c20153988fdf76e7f974fd`

Both child PRs are merged with zero unresolved review threads. The 365Bench exact-head SonarCloud gate passes.

## Summary by Sourcery

Enhancements:
- Advance the 365Bench and cloudBench subrepositories to their latest merged heads.

PR #19, opened 2026-09-01
Advance merged pyBench gitconfig startup fix

brettheap:
## Summary

Advance the pyBench submodule to merged pyBench PR #2, which validates the host `.gitconfig` mount before startup and preserves explicit Docker Compose failure reporting.

## Validation

- pyBench `bash -n` and ShellCheck
- pyBench Compose contract test
- missing, empty-directory, and non-empty-directory `.gitconfig` safety probes
- exact child merge commit: `8e164196dad7952217b3dc5f9907bddbd49bef34`

SonarQube analysis remains advisory because the licensed LOC ceiling may prevent it from running.

## Summary by Sourcery

Enhancements:
- Advance the pyBench submodule to the merged startup-safety fix for validating host `.gitconfig` mounts and preserving Docker Compose failure reporting.

PR #59, opened 2026-09-11
Re-vendor openRepoTools at a40cbb7: `status` reads a record's `pushed:` the way `resume` will judge it — unreadable, or missing altogether, is neither true nor false, and neither is sent to `resume`

brettheap:
The `openrepotools` source pin moves from a2c85219942eb8a8fbb2cfd20131e0ebf0336c10
(pinned in #54) to a40cbb7403fcc730c4590e8bd284196d0a91f9f0 —
`gh api repos/opensoft/openRepoTools/compare/main...a40cbb7403fcc730c4590e8bd284196d0a91f9f0`
says `identical`, so it is the commit opensoft/openRepoTools's main carries after
opensoft/openRepoTools#20.

Between the two commits (opensoft/openRepoTools#19 and #20, both squash-merged),
only `status` changes, in its parked-record layer, and both are follow-ups to
reviewer notes rather than rulings: a record's `pushed:` is read the way the
Speckit git extension's `resume.sh` will judge it. A value that is neither `true`
nor `false` — a bare `pushed:`, `pushed: ""`, `pushed: yes` — used to fall through
every `= false` test and read as pushed, so a hand-edited record drew "the record
says it was pushed" and, with no worktree here, was sent to `resume <Name>`, which
refuses such a leg (`resume.sh`'s `[ "$pushed" != true ]`); it is now named for
what it is, in a finding of its own, with the exit that works — park it again from
the workstation that parked it — and a trailing `#` comment is stripped from the
value as the extension's loader strips it from every line (#19). And a leg with no
`pushed:` KEY at all, which the record layer used to drop without a word while the
extension's loader defaults it to `false` and `resume` refuses it in the
`--no-push` words, is a fourth state of the field: the leg's row now goes out where
the leg ends rather than at its `pushed:` line, a sixth field tells absent from
valueless, and the finding says what the record does not say, what `resume` makes
of that, and the same exit; a full-line `#` comment inside a leg is no leg-ender
(#20). `status --help` carries "missing or neither true nor false" beside the
`--no-push` exception. `park`, `resume` and the installer are byte-identical to the
#54 vendoring.

So this is one `apply` run and nothing else: no command was added or removed, so
the Dockerfile COPYs, scripts/setup-estate-commands.sh, setup.sh, the README and the
test scripts that enumerate the commands are untouched. update-upstream.py is untouched.

devBenches/devcontainer.test/test-speckit-git-feature.sh: 56 PASS, GREEN, exit 0.
The base image was not rebuilt here; only the vendored copy of `status` and its pin rows change.

## Summary by Sourcery

Re-vendor openRepoTools to make `status` consistently interpret parked-record push state with `resume`.

Bug Fixes:
- Align `status` handling of missing, valueless, and invalid `pushed:` fields with the behavior used by `resume`, including clear findings and recovery guidance.

Enhancements:
- Re-vendor openRepoTools `status` from commit a40cbb7 with improved parked-record parsing and reporting.

Documentation:
- Clarify in `status --help` how missing or non-boolean `pushed:` values are handled for `--no-push` records.

Chores:
- Update the vendored source pin and checksum for openRepoTools.

### DIFF
--- a/devBenches/base-image/files/openrepotools/status
+++ b/devBenches/base-image/files/openrepotools/status
@@ -144,13 +144,15 @@ pin that has fallen behind is named with the exit, `update-shape.py check
 --root <root>`. THE PARKED RECORD (`~/.agents/workspace.yaml` names the
 workspace repository; `workspaces/<org>/<id>.yaml` in it) is read against the
 worktrees on disk: a recorded feature with no worktree here (`resume <Name>`
-brings it back — unless the record says `--no-push`, when only the
-workstation that has the WIP commit can park it again; and where `git
-worktree list` still holds a registration that is no longer a worktree, that
-is cleared with `worktree prune`, after a `worktree unlock` if it is locked
-and with any leftover directory moved aside, before `resume` can recreate
-anything), a worktree whose tip is not the parked commit, a leg parked with
-`--no-push`, and a worktree the record does not know (never parked).
+brings it back — unless the record says `--no-push`, or its `pushed:` is
+missing or neither true nor false, which `resume` refuses the same way: then
+only the workstation that parked it can park it again; and where `git worktree
+list` still holds a registration that is no longer a worktree, that is cleared
+with `worktree prune`, after a `worktree unlock` if it is locked and with any
+leftover directory moved aside, before `resume` can recreate anything), a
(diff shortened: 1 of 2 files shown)


brettheap:
@codex review

PR #60, opened 2026-09-11
resume.sh's RR6 says what the record's `pushed:` says: only `false` is the `--no-push` park, and a value that is neither true nor false is named, not read as a claim nobody made

brettheap:
A follow-up to a reviewer's note recorded on opensoft/openRepoTools #19 and #20 (2026-09-11), taken on the note alone — nobody has ruled on it, and it is yours to decide. `resume.sh`'s RR6 refused every leg whose `pushed:` was not `true` in the words "was parked with --no-push", so a record whose `pushed:` said `maybe`, had no value, or was missing altogether drew a `--no-push` claim nobody made — and, because `workspace_load_project` seeded a missing key as `false` and `park.sh`'s `emit_recorded_feature` carries a REFUSED feature's loaded entry forward, a park that refused such a feature then WROTE `pushed: false` into the record, persisting the claim. Three commits, each droppable on its own:

1. RR6 says what the record says. The refusal is exactly as strict (`!= true` still refuses every non-`true` value, nothing is recreated), but `false` alone is the `--no-push` park — its three stderr lines, the REFUSED summary and the `--json` reason are byte-identical to today's — while a value that is neither true nor false is named for what it is: "its record's `pushed:` says 'maybe', which is neither true nor false", "… has no value, …", "the record has no `pushed:` for this leg", each followed by "Whether its parked commit … ever left <workstation> cannot be read from the record." and "Park it again from there, which writes the record afresh, then resume here." — the same story openRepoTools' `status` now tells for the same record. Two rows in `commands/speckit.git.resume.md`.
2. The loader stops inventing `false`. `workspace_load_project` gains `MANIFEST_LEG_PUSHED_SEEN[]` (set where a `pushed:` key was read) and seeds an absent key as the empty string, so absent and bare stay apart; `MANIFEST_LEG_PUSHED` has exactly two readers (`resume.sh`'s RR6 and `park.sh`'s `emit_recorded_feature`), both tolerant of the empty string, and `workspace_write_manifest` already drops an empty value — so a refused park now carries the hole forward instead of writing `pushed: false`, while `park`'s own success path still writes the literal `true`/`false` it computes for every leg it parks. Every record any past `park` wrote carries `pushed:` for every leg, so nothing already on disk reads differently. There is no PowerShell twin of park/resume/the loader to move in step.
3. From the independent review of this branch: the branch-level REFUSED line and the `--json` reason for an absent key name the leg ("for the spec leg") rather than "this leg", which had no antecedent there; and the absent-key test asserts the refused park really rewrote the manifest before asserting what it did not write.

Proved: `devBenches/devcontainer.test/test-speckit-git-feature.sh` 59 PASS / 0 FAIL (56 on main; the two biting scenarios fail against main's scripts with exactly the old `--no-push` text, and the write-back half fails on the manifest — `pushed: false` invented); `test-speckit-git-compat.sh` 22 / 0; both suites under the workflow's pinned bash 3.2.57 and git 2.34.1 images, 81 / 0 each; `update-upstream.py check` 6 vendored copies match (nothing vendored is touched); shellcheck findings identical to main's on all three scripts; no bash-4 construct. The independent review (168 runs of resume.sh across eight `pushed:` spellings, main against this branch; a six-leg two-project manifest through the loader proving the parallel arrays stay in step; the write-back proved on a multi-leg fixture both ways) returned "merge as is".

Follow-up this creates, in openRepoTools and then a re-vendor here: `status`'s absent-`pushed:` clause says "`resume` reads a missing one as not pushed and refuses it" — after commit 2 the reason is stale (the outcome, refused and re-park, is unchanged) — and three comments there describe the old loader; that change lands in openRepoTools first, then the vendored `status` moves here with a pin bump. The vendored copy is deliberately untouched in this PR.

Pre-existing, unrelated: `devcontainer.test/test-openspeckit-bootstrap.sh` has two failing lines on main today (OPSX template hash), byte-identical before and after this branch.

## Summary by Sourcery

Make resume report only what the manifest establishes about whether parked work was pushed, while preserving strict refusal for uncertain records.

New Features:
- Distinguish explicit `pushed: false`, invalid values, bare values, and missing keys when reporting resume refusals.

Bug Fixes:
- Prevent resume from falsely claiming a feature was parked with `--no-push` when the manifest does not make that claim.
- Preserve missing `pushed:` fields when a refused park rewrites the workspace manifest instead of inventing `pushed: false`.

Enhancements:
- Keep manifest loading aware of whether each leg explicitly contained a `pushed:` key while retaining strict refusal of every value other than `true`.
- Align resume refusal messages and summaries with the information actually recorded in the manifest.

Documentation:
- Document the separate handling of invalid and missing `pushed:` metadata in the resume command guidance.

Tests:
- Add coverage for explicit no-push records, invalid and valueless `pushed:` entries, and missing keys including refused-park write-back behavior.

### DIFF
--- a/devBenches/base-image/files/speckit-worktree/templates/specify/extensions/git/commands/speckit.git.resume.md
+++ b/devBenches/base-image/files/speckit-worktree/templates/specify/extensions/git/commands/speckit.git.resume.md
@@ -66,6 +66,7 @@ Print the script's own message verbatim, then add the reading below.
 | the local branch is BEHIND the parked commit | This workstation had the feature before the other one parked newer work. The remediation is the fast-forward the message prints, then `make resume` again — not a deletion |
 | the local branch diverged | Local commits exist that the parked commit does not contain. Rename or delete that branch; look first with the printed `log` command |
 | parked with `--no-push` | The parked commit exists only on the workstation that parked it. Push it there, re-run `park`, then resume here |
+| the record's `pushed:` is neither true nor false | A value `park` never writes — a hand-edit or a bad merge of the record. Do NOT read it as `--no-push`: the record rules nothing out, so whether the parked commit ever left that workstation cannot be read from it. The exit is a park from the workstation that has the worktree, which writes the record afresh |
 | the workspace has no entry for this project | Either nothing was parked for it, or it was parked into a differently named file; the message prints the `grep` over that org's directory |
 | `workspace-config-invalid` | The config's `orgs:` block is malformed. It refuses rather than falling back to the default workspace, because a confidential org's work must never be indexed in the wrong repository |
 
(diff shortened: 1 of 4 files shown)


brettheap:
@codex review

PR #15, opened 2026-09-01
Add governed pyBench SonarQube MCP change

brettheap:
## Summary

- add the OpenSpec/Speckit governance scaffold used by the pyBench SonarQube MCP change
- document the governed container-runtime routing change
- preserve the previously local-only commit as a reviewable stacked branch

## Stack

- base: `main` (workBenches PR #3 is merged)
- follow-up: `001-pybench-sonarqube-mcp`

## Validation

- generated workflow scaffold plus focused fixes for portable config parsing and worktree selection
- SonarQube is advisory because the current licensed LOC ceiling may prevent analysis

PR #63, opened 2026-09-12
claude-profile: the lane defaults from the tmux window, else the swap record (Amendment 8(c))

brettheap:
Lane: openrepoproject-1 (openRepoProject-1)
Claim: https://github.com/brettheap/new-workstation/issues/15#issuecomment-5647279137

The launcher half of **Lane Collision Protocol Amendment 8**, drafted against the SPEC comment on brettheap/new-workstation#15 (clause (c), "a restart is one launch"). **This lands after the opensoft/brett-wip PR**, which adds the two things it codes against: `lanes-edit.sh swapped` and `lane-start --confirm`, and `lanes-edit.sh session-start` for the hook below. Neither exists on brett-wip `main` today, so every call to them here is guarded and degrades to today's behaviour — merging this first changes nothing for anyone except the note in case 4, but it is not *useful* until the other half lands.

## What was wrong

`--lane` was opt-in. A launch without it started a Claude with no lane at all: the window kept tmux's automatic rename, no register row was stamped, and the resume fell to `/resume`'s picker filtered by a title that several transcripts in one directory share. That is the 2026-09-11 wrong-lineage resume, whose sixth finding is, verbatim, "**The automation never ran.** The launcher (`claude-profile`) was invoked without `--lane`" (lane-collision-protocol Amendment 6). Typing `--lane <repo>-<n>` correctly, once per window, thirteen times in a day of usage resets, is the part that does not survive contact with a swap.

## What lands

`base-image/files/claude-profile` — the file behind `~/.local/bin/pclaude` and `~/.local/bin/claude-profile` — resolves the lane itself for a `run` that starts a conversation:

| # | where the lane comes from | handed to `lane-start` as |
|---|---|---|
| 1 | `--lane` / `CLAUDE_LANE` | *bare* |
| 2 | the CURRENT tmux window's name, when `lanes-edit.sh register-row <name>` answers 0 | *bare* |
| 3 | `lanes-edit.sh swapped <ws>`, first row (most recent), **inside tmux only** | `--confirm` |
| 4 | nothing | *nothing* — today's behaviour, plus one line: `pclaude: no lane for this window; run lane-start <repo> <n> inside it`, and not even that where the clause (e) hook below is ensured |

2 is certain: the operator is standing in the lane's own window, and `lane-start` turned tmux's automatic rename off when it took that window, so a window still called `claude` or `zsh` simply is not a lane and only exit 0 says otherwise. 3 is a guess read from a record written *before* the restart, so `lane-start` is asked to confirm it before it renames the window — and it is a guess about a *window*, which is why it is not read at all outside tmux, where there is no window to rename.

**`--yes` is passed nowhere** (F-W2, F-T8): it is inert in `lane-start` unless `--confirm` came with it, and `--confirm` is the one case it must never come with, since answering the question in advance is the opposite of asking it. Cases 1 and 2 go over bare — byte-for-byte what `--lane` did before this default existed.

`--no-lane` (or `CLAUDE_NO_LANE=1`) opts out entirely, and wins over `--lane`/`CLAUDE_LANE` in either order.

**The window name is captured before the re-exec and carried across it** as `WORKBENCHES_CLAUDE_WINDOW`, the same explicit threading `CLAUDE_LANE` already gets. The re-exec into a new tmux session is otherwise untouched.

`lanes-edit.sh` is found the way `lane-start` is, and never as a path spelled for one workstation: `$WORKBENCHES_LANES_EDIT_BIN`, else on `PATH`, else beside the `lane-start` that IS on `PATH` (both are `lanes/` in opensoft/brett-wip), else the `~/projects/xFactory/lanes-edit.sh` symlink the protocol names.

## Every way it degrades, and what happens

| condition | behaviour |
|---|---|
| `lanes-edit.sh swapped` unknown (exit 2 — brett-wip `main` today) | case 4 |
| `swapped` fails after printing a row | case 4 — output from a read that did not succeed is not a lane |
| a `swapped` row not tab-separated, or whose first field is not lane-shaped | case 4 — the tab is the contract, and `cut -f 1` would hand back a whole line that has none |
| `swapped` exits 2 **and says "unknown subcommand"** | case 4, silently — an old helper |
| `swapped` exits 2 and does **not** say that | case 4, with one line naming it a caller bug: 2 is also `lanes-edit.sh`'s usage code |
| the launch is outside tmux (`WORKBENCHES_CLAUDE_TMUX=off`, no tty) | case 4 — a record lane needs a window to take |
| `lane-start --confirm` is answered **`N`** | Claude launches bare, exit 0, one line — the launcher runs `lane-start` on that path instead of `exec`ing it, so it is still there to catch the 2 |
| `lane-start` took the lane and its Claude exited 2 | that status is handed back; the window name says the lane was taken, so no second Claude is started |
| `register-row` answers 8 (no row) or 2 (not a lane-shaped window name) | fall through to 3 |
| `lane-start --help` offers no `--confirm` (brett-wip `main` today), lane from the **record** | **dropped**, with the reason, then case 4 |
| no `lanes-edit.sh` on the machine | case 4 |
| no `lane-start` on `PATH` — no lane estate at all | silent; not even the note |
| `mcp`, `plugin`, `doctor`, `install`, `update`, `--help`, `--version`, `--print`, `-p` | no resolution, no note — not a conversation |

**Nothing in the resolution can refuse a launch** — and since A8 Addendum 2 R-A8-3 that is true of the composition and not just of this file. The blocking finding F-W1 was that `exec`ing `lane-start` made *its* status the launcher's, and `lane-start` exits 2 when `--confirm` is answered `N`: the operator who said "no, not that lane" got an exited pane and no Claude. One question, and both of its answers now lead somewhere.

## The other half of R-A8-5: the hook and the skill reach the profiles the launcher actually uses

Every `pclaude run` execs with `CLAUDE_CONFIG_DIR=<profile dir>`, so the harness reads **that** directory's `settings.json` and **that** directory's `skills/` — never `~/.claude`'s. Measured on Eagle: 418 profile directories, none with a `SessionStart` hook, each `skills` a symlink to one shared directory that was empty. Amendment 8's hook had therefore never fired and its skill had never been listed (F-S1, F-S2).

- **The hook.** `configure_profile_runtime` ensures clause (e)'s entry in the profile's own `settings.json` on every run, matched by the **exact command string** and idempotent on it: an entry already carrying that command is left exactly as it is — its matcher, its timeout, whatever else sits beside it — and every other key under `.hooks` is untouched, exactly as the `UserPromptSubmit` rewrite beside it already was. The entry is `{"matcher": "startup|resume|clear|fork", "hooks": [{"type": "command", "command": "~/projects/xFactory/lanes-edit.sh session-start || true", "timeout": 5}]}`. It is written only where `~/projects/xFactory/lanes-edit.sh` exists **and** has the `session-start` subcommand: `|| true` hides the status of a helper that has none, not the usage block it prints, and that block would otherwise stand in front of every session start on the workstation until the brett-wip half lands. A later install is picked up by the next launch, because the ensure runs on every one.
- **The skill.** `/lane-swap` is vendored at `base-image/files/claude/skills/lane-swap/SKILL.md` and installed by `scripts/setup-claude-profiles.sh` into `~/.claude-profiles/shared/skills/` — the one directory all 418 profiles' `skills` symlink already points at — and into `~/.claude/skills` for a bare `claude`. The copy is idempotent by content. **Those are the same paths Amendment 9 assigns to `openRepoTools --install`: ruled on opensoft/workBenches#68 (F5), this loop stands for Amendment 8 and is their sole writer until A9's adoption act 3 lands, and A9's act 4 deletes it.**
- **One no-lane notice, not two.** Clause (e) makes the hook's block the only no-lane notice in the estate and stands clause (c) step 4 down wherever the hook can speak. The launcher is what ensures that hook, so it knows when it will speak, and says nothing when it will.

**Live state (RV-W3, re-verification round), so the tables above are never read as a live measurement.** Nothing here has fired under a profile launch yet. Measured on Eagle just now: 418 profile `settings.json` under `~/.claude-profiles/profiles/*/*/*/`, 0 carrying a `SessionStart` hook; `~/.local/bin/pclaude` still resolves to the live `workBenches` checkout, not this branch; and `~/projects/xFactory/lanes-edit.sh` -> `brett-wip/lanes/lanes-edit.sh` on `main`, which has 0 occurrences of `session-start`. Both writers of the entry (this launcher, per-profile, and `scripts/setup-claude-profiles.sh`, for the bare-`claude` path — added this round, RV-W2) are therefore correctly ensuring nothing today. This section and the one below it describe the code path once the brett-wip half of Amendment 8 lands, not a live measurement of any workstation.

## Two things in the spec's Launcher section that the launcher's structure makes wrong, reported rather than worked around

**1. "Capture the window name before the re-exec and carry it across" is unreachable today — and the code that matters is the other half of it.** `claude_run_is_interactive` re-execs into a new tmux session only when `TMUX` is *empty*: a `pclaude` typed inside a tmux window never creates a session, it runs in the window it was typed in. So the two cases never overlap — when there is a window name to carry there is no re-exec, and when there is a re-exec there was no window — and the carried value is always empty in practice. I implemented the carry anyway, because what the child-side code is really for is the opposite direction: without it the child would ask tmux for `#W` and get `claude-team-002-20260912170400-1234`, the brand-new session's own window name, and any `register-row` that answered 0 for such a name would be a lane taken from a window nobody named. The child now reads only what was carried, and asks tmux nothing. Both halves are pinned by scenario 10 of the test, including the assertion that a re-exec carries no window name.

**2. A record lane is dropped, not silently taken, when `lane-start` cannot ask.** The spec says the record's lane goes over with `--confirm`; it does not say what to do while `--confirm` does not exist. Handing the lane over bare would take a window on a guess with no question asked, which is the one thing clause (c) exists to prevent — so the launcher declines the guess, says why on stderr, and prints the note. A lane from the *window* is still handed over bare in that case, since nothing is being guessed and that is byte-for-byte today's behaviour.

Two further things left alone, for your call:

- **`register-row` fetches.** It calls `log_sync`, which runs `git fetch origin` — a network round-trip on the path of every launch, and a hang in the one place a hang is least acceptable. Both reads are therefore made with `LANES_NO_FETCH=1`: a row or a record *this* workstation wrote is in this workstation's own clone already. If a row published from ANOTHER workstation must be resolvable here, that read has to be a real fetch, and a launcher is the wrong place to put one.
- **Amendment 6's open item is untouched.** A lane whose checkout is not `$PROJECTS_ROOT/<repo>` still cannot be launched by lane name, resolved or typed; `lane-start` derives the directory from the repo name and the launcher passes only the name. The default merely means the launcher now reaches that limit without being asked to. `--no-lane` is the way out until it is closed.

## How it was proven

**New:** `devcontainer.test/test-claude-profile-lane-default.sh`, standalone like its sibling `test-claude-profile-lane-start.sh` (which `test.sh` does not wire in either), taking the launcher path as `$1`. It fakes `tmux` (a settable `#W`), `lanes-edit.sh` (`register-row` 0/8/2, `swapped` 0/8/2, and a `swapped` that prints and *then* fails), `lane-start` (logs its argv; answers `--help` with this scenario's help) and `claude`, pins `PATH` and `HOME` so the workstation's real lane estate cannot answer, and covers all four cases, both flag choices, both opt-outs, the tmux-child carry, and every row in the degrade table. It also asserts both register reads carry `LANES_NO_FETCH=1`.

Two more suites came with the fix round: `devcontainer.test/test-claude-profile-session-start-hook.sh` (the hook entry: ensured, idempotent, preserved when present, additive beside other hook kinds, not written where the estate cannot answer it) and `devcontainer.test/test-claude-profile-skill-install.sh` (the installer, and the skill's own text, including that every fenced shell block in it parses). The first two are registered in `devcontainer.test/test.sh`; the third is host-run, like `test-setup-estate-commands.sh`, because it needs `scripts/` and `base-image/`.

**It bites. Thirty-nine mutations, each one edit a careless change could make, each caught** — the original sixteen (record-before-window; always `--yes`; `--confirm` for a window lane; any `register-row` exit treated as a lane; `swapped`'s exit status ignored; the lane-shape check dropped; the whole row taken as the lane; the last row taken instead of the first; the `--confirm` degrade removed; the `--no-lane` opt-out removed; an explicit lane not winning; the not-a-conversation gate removed; the tmux-child guard removed; `LANES_NO_FETCH` dropped; the note removed; the note printed with no lane estate) and twenty-three more from this round: the confirm path `exec`ing again, the fall-through firing on any status, its window check dropped, the caller-bug branch removed and over-firing, the tab check dropped, the tmux gate removed, the note standing down always and never, the hook's command string varied, its timeout dropped, its matcher changed, its idempotence key broken two ways, its estate gate dropped two ways, the `UserPromptSubmit` rewrite made destructive, the skill's installer made non-idempotent and skipping the shared directory, and four mutations of the skill's own text.

Three of them escaped the first run, and each escape was a real hole — one of which was a scenario that had never tested what it claimed: its fake `lanes-edit.sh` selected the old-helper branch with `break` inside a `case`, which is meaningless outside a loop, so bash warned and ran the other arm.

**Unchanged and still green,** all on this branch's HEAD, on the host:

| run | result |
|---|---|
| `devcontainer.test/test-claude-profile-lane-default.sh` | 31 scenarios, 106 assertions |
| `devcontainer.test/test-claude-profile-session-start-hook.sh` | 10 scenarios, 25 assertions (new) |
| `devcontainer.test/test-claude-profile-skill-install.sh` | 4 scenarios, 34 assertions (new) |
| `devcontainer.test/test-claude-profile-lane-start.sh` | passes, untouched — including its byte-for-byte "no lane named" baseline |
| `devcontainer.test/test-claude-profile-preserves-model.sh` | passes |
| `devcontainer.test/test-claude-tmux-statusline.sh` | passes |
| `devcontainer.test/test-claude-tmux-statusline-env-isolation.sh` | passes |
| `bash -n` on both files | clean |
| `koalaman/shellcheck:stable -s bash` on both files | run in the first round only; no shellcheck on this host in the fix round |

Not proven here: the container suite (`devcontainer.test/test.sh`) was not built and run — the two lane suites are registered in it but the image was not rebuilt — and neither the real `lanes-edit.sh swapped` nor the real `lane-start --confirm` nor `lanes-edit.sh session-start` exists yet to run against. That pairing is the brett-wip PR's proof and this repository's next one.

`docs/claude-multi-account-profiles.md` moves with the behaviour: the paragraph that said `--lane` was "opt-in, and off by default" now describes the resolution, both opt-outs, and the degrades.

## Summary by Sourcery

Make Claude profile launches automatically preserve and recover lane context while adding the supporting swap skill, session hook, documentation, and regression coverage.

New Features:
- Automatically resolve lanes for conversation launches from explicit settings, the current tmux window, or the latest swap record with safe fallback behavior.
- Add the `/lane-swap` skill and install it for shared Claude profiles and bare Claude sessions.
- Ensure Amendment 8 SessionStart lane hooks in profile and default Claude configuration when the lane estate supports them.

Bug Fixes:
- Prevent launches from losing lane context when `--lane` is omitted, including after swaps and tmux re-exec.
- Avoid failed confirmations, unavailable capabilities, malformed records, and declined lane claims from preventing Claude from starting or launching a duplicate session.

Enhancements:
- Make lane resolution opt-out with `--no-lane` or `CLAUDE_NO_LANE`, preserve explicit lane precedence, and avoid network fetches during launch-time lookups.
- Keep hook configuration additive and idempotent while preserving existing custom entries.
- Update multi-account profile documentation to describe automatic lane resolution, hooks, skills, and fallback behavior.

CI:
- Add regression coverage for lane resolution, SessionStart hook management, skill installation, degradation paths, tmux carry-over, and profile configuration behavior.

Documentation:
- Document the Amendment 8 lane default, SessionStart hook, shared skill installation, and bare-Claude behavior.

Tests:
- Add standalone and devcontainer test suites covering 31 lane-resolution scenarios, hook configuration, skill installation, shell parsing, and idempotence.

### DIFF
--- a/base-image/files/claude-profile
+++ b/base-image/files/claude-profile
@@ -132,18 +132,25 @@ claude_add_has_scope() {
   return 1
 }
 
-claude_run_is_interactive() {
-  [[ -t 0 && -t 1 ]] || return 1
-  [[ "${WORKBENCHES_CLAUDE_TMUX:-auto}" != off ]] || return 1
-  [[ -z "${WORKBENCHES_CLAUDE_TMUX_CHILD:-}" ]] || return 1
-  [[ -z "${TMUX:-}" ]] || return 1
-
+# The first Claude argument says whether this launch is a conversation at all.
+# `mcp`, `doctor`, `--version` and the other subcommands and one-shots are not,
+# so they are neither wrapped in a tmux session of their own nor handed to
+# lane-start by the lane default below.
+claude_args_start_session() {
   case "${1:-}" in
     mcp|plugin|plugins|doctor|install|update|--help|-h|--version|-v|--print|-p)
       return 1
       ;;
   esac
   return 0
+}
+
+claude_run_is_interactive() {
+  [[ -t 0 && -t 1 ]] || return 1
+  [[ "${WORKBENCHES_CLAUDE_TMUX:-auto}" != off ]] || return 1
+  [[ -z "${WORKBENCHES_CLAUDE_TMUX_CHILD:-}" ]] || return 1
+  [[ -z "${TMUX:-}" ]] || return 1
+  claude_args_start_session "$@"
 }
 
 start_profile_tmux() {
@@ -157,6 +164,12 @@ start_profile_tmux() {
   # on an already-running server, so it is threaded through explicitly here
   # rather than relied on via inheritance — same idiom as the guard above.
   [[ -n "${lane:-}" ]] && env_prefix+=("CLAUDE_LANE=$lane")
+  [[ "${no_lane:-false}" == true ]] && env_prefix+=("CLAUDE_NO_LANE=1")
+  # So must the name of the window this was typed in. The session created here
+  # is brand new and its window is named for the command that made it, never
+  # for a lane, so the child must read the name captured on this side of the
+  # re-exec instead of asking tmux again (capture_lane_window).
+  [[ -n "${lane_window:-}" ]] && env_prefix+=("WORKBENCHES_CLAUDE_WINDOW=$lane_window")
 
   command -v tmux >/dev/null 2>&1 || {
     echo "pclaude: tmux is required for interactive profile launches." >&2
@@ -251,19 +264,158 @@ profile_metadata() {
   return 1
 }
 
-# --lane / CLAUDE_LANE (opt-in, default off): hand the final Claude launch to
-# `lane-start` instead of exec'ing Claude directly. Recognised only as a
-# leading option, before the action/profile — the first token that is not
-# --lane/--lane=.../-h/--help ends this loop and is left for the existing
-# parsing below, so anything meant for Claude itself (e.g. `claude-profile
-# work --help`) is untouched.
+# ---------------------------------------------------------------------------
+# THE LANE DEFAULT — lane-collision-protocol Amendment 8(c).
+#
+# `--lane` was opt-in, and a launch without it started a Claude with no lane at
+# all: the window kept tmux's automatic rename, no register row was stamped,
+# and the resume was left to a picker filtered by a title several transcripts
+# share. That is the 2026-09-11 wrong-lineage resume, whose sixth finding was
+# simply "the automation never ran" (Amendment 6). So a launch that starts a
+# conversation now takes a lane by default, and the lane is RESOLVED rather
+# than typed:
+#
+#   1. --lane / CLAUDE_LANE — what the operator said. Always wins.
+#   2. the CURRENT tmux window's name, when `lanes-edit.sh register-row` says
+#      the register has a row for it. The window name IS the lane (Rule 4),
+#      and lane-start turned tmux's automatic rename off when it took the
+#      window, so a window still called `claude` or `zsh` simply is not one.
+#   3. `lanes-edit.sh swapped <workstation>` — the lane this workstation last
+#      PAUSED for a swap and has not resumed since (the record, Amendment
+#      8(b)). Its first row is the most recent.
+#   4. nothing: exactly today's behaviour, plus one line saying how to take a
+#      lane in this window.
+#
+# 2 is certain — the operator is standing in the lane's own window — so it goes
+# to lane-start as `--yes`, as an explicit `--lane` does. 3 is a guess read
+# from a record written before the restart, so it goes as `--confirm` and
+# lane-start asks before it renames the window.
+#
+# Nothing here refuses a launch. Every probe that cannot be made — no
+# lane-start, no lanes-edit.sh, a lanes-edit.sh with no `swapped` yet, a
+# lane-start with no `--confirm` yet — falls back to 4.
+
+lanes_edit_bin() {
+  # Found the way lane-start is, and never as a path spelled for one
+  # workstation: on PATH, else beside the lane-start that IS on PATH (both are
+  # `lanes/` in opensoft/brett-wip), else the symlink the protocol names.
+  local candidate lane_start
+  candidate="${WORKBENCHES_LANES_EDIT_BIN:-$(command -v lanes-edit.sh 2>/dev/null || true)}"
+  if [[ -z "$candidate" ]]; then
+    lane_start="$(command -v lane-start 2>/dev/null || true)"
+    if [[ -n "$lane_start" ]]; then
+      candidate="$(dirname "$(readlink -f "$lane_start")")/lanes-edit.sh"
+    fi
+  fi
+  [[ -n "$candidate" ]] || candidate="$HOME/projects/xFactory/lanes-edit.sh"
+  [[ -x "$candidate" ]] || return 1
+  printf '%s\n' "$candidate"
+}
+
+capture_lane_window() {
+  # The lane's window is the one this was typed in, and its name is readable
+  # only HERE — before start_profile_tmux re-execs into a session tmux names
+  # after the command that made it. A child of that re-exec never asks tmux: it
+  # takes the name threaded across, which is empty when the launch did not come
+  # from a tmux window at all.
+  if [[ -n "${WORKBENCHES_CLAUDE_TMUX_CHILD:-}" ]]; then
+    lane_window="${WORKBENCHES_CLAUDE_WINDOW:-}"
+    return 0
+  fi
+  lane_window=""
+  [[ -n "${TMUX:-}" ]] || return 0
+  command -v tmux >/dev/null 2>&1 || return 0
+  lane_window="$(tmux display-message -p '#W' 2>/dev/null || true)"
+  return 0
+}
+
+lane_note_no_lane() {
+  echo "pclaude: no lane for this window; run lane-start <repo> <n> inside it" >&2
+}
+
+lane_start_supports() {
+  # `--confirm` and `--yes` arrive with the brett-wip half of Amendment 8.
+  # Until they do, a lane-start that has never heard of them must not be handed
+  # one, and its own --help is the only thing that can say.
+  local flag="$1"
+  if [[ -z "${lane_start_help+set}" ]]; then
+    lane_start_help="$(lane-start --help 2>/dev/null || true)"
+  fi
+  [[ "$lane_start_help" == *"$flag"* ]]
+}
+
+resolve_lane() {
+  # Sets `lane` and `lane_source` (flag|window|record, empty for none) from the
+  # Claude arguments this launch carries. Never fails, never refuses: a lane
+  # that cannot be resolved is today's behaviour.
+  local lanes_edit ws rows status candidate
+  local -a swapped_argv
+  if [[ -n "${lane:-}" ]]; then
+    lane_source=flag
+    return 0
+  fi
+  [[ "$no_lane" == false ]] || return 0
+  claude_args_start_session "$@" || return 0
+  # No lane estate on this machine: the whole feature is absent, and saying so
+  # on every launch would be noise.
+  command -v lane-start >/dev/null 2>&1 || return 0
+  lanes_edit="$(lanes_edit_bin || true)"
+  if [[ -z "$lanes_edit" ]]; then
+    lane_note_no_lane
+    return 0
+  fi
+
+  # LANES_NO_FETCH=1 on both reads: they are on the path of every launch, and
+  # lanes-edit.sh otherwise fetches origin before it answers. A row or a record
+  # this workstation wrote is in this workstation's own clone already, and a
+  # launcher must not sit on the network to open a terminal.
+  #
+  # Only 0 is a lane. `register-row` answers 8 for a name with no row and 2 for
+  # a window name that is not even lane-shaped (`zsh`, `1:vim*`).
+  if [[ -n "${lane_window:-}" ]] \
+    && LANES_NO_FETCH=1 "$lanes_edit" register-row "$lane_window" >/dev/null 2>&1; then
+    lane="$lane_window"
+    lane_source=window
+    return 0
+  fi
+
+  ws="${LANES_WORKSTATION:-$(hostname -s 2>/dev/null || hostname 2>/dev/null || true)}"
+  swapped_argv=(swapped)
+  [[ -z "$ws" ]] || swapped_argv+=("$ws")
+  set +e
+  rows="$(LANES_NO_FETCH=1 "$lanes_edit" "${swapped_argv[@]}" 2>/dev/null)"
+  status=$?
+  set -e
+  # 0 with rows `<lane>\t<UTC>\t<window>`, most recent first; 8 with none; 2
+  # from a lanes-edit.sh that has no `swapped` subcommand yet.
+  if [[ "$status" -eq 0 && -n "$rows" ]]; then
+    candidate="$(printf '%s\n' "$rows" | head -n 1 | cut -f 1)"
+    if [[ "$candidate" =~ ^[A-Za-z0-9_][A-Za-z0-9._-]*$ ]]; then
+      lane="$candidate"
+      lane_source=record
+      return 0
+    fi
+  fi
+  lane_note_no_lane
+  return 0
+}
+
+# --lane / CLAUDE_LANE / --no-lane are recognised only as leading options,
+# before the action/profile — the first token that is not one of them ends this
+# loop and is left for the existing parsing below, so anything meant for Claude
+# itself (e.g. `claude-profile work --help`) is untouched.
 lane="${CLAUDE_LANE:-}"
+lane_source=""
+lane_window=""
+no_lane=false
+[[ -z "${CLAUDE_NO_LANE:-}" ]] || no_lane=true
 while [[ $# -gt 0 ]]; do
   case "$1" in
     --lane)
       [[ $# -ge 2 ]] || { echo "pclaude: --lane needs a value: <repo>-<n>" >&2; exit 2; }
       lane="$2"; shift 2 ;;
     --lane=*) lane="${1#--lane=}"; shift ;;
+    --no-lane) no_lane=true; shift ;;
     -h|--help)
       cat <<'EOF'
 Usage: claude-profile {list|login|status|run} PROFILE [CLAUDE_ARGS...]
@@ -277,7 +429,7 @@ Actions:
                         "run" itself may be omitted: claude-profile PROFILE)
 
 Options:
-  --lane <repo>-<n>    opt-in: hand the final launch to
+  --lane <repo>-<n>    take this lane: hand the final launch to
                         `lane-start <repo>-<n> -- CLAUDE_ARGS` instead of
                         exec'ing Claude directly (also settable as
                         CLAUDE_LANE=<repo>-<n> in the environment).
@@ -286,14 +438,27 @@ Options:
                         with --resume/--name <lane> under this same profile.
                         Requires lane-start on PATH (opensoft/brett-wip,
                         lanes/, linked by that repo's scripts/link-estates).
-                        Without --lane/CLAUDE_LANE, behaviour is unchanged.
+  --no-lane            take no lane: exec Claude directly, whatever the
+                        window or the swap record says (also settable as
+                        CLAUDE_NO_LANE=1 in the environment). Wins over
+                        --lane/CLAUDE_LANE.
   -h, --help           show this help
+
+THE LANE DEFAULT (lane-collision-protocol Amendment 8(c)). With no --lane, a
+`run` that starts a conversation resolves the lane itself: the current tmux
+window's name when the lane register has a row for it (handed over as --yes),
+else the lane this workstation last paused for a swap and has not resumed
+since (handed over as --confirm, so lane-start asks before it takes the
+window). With neither, and with --no-lane, behaviour is unchanged.
 EOF
       exit 0
       ;;
     *) break ;;
   esac
 done
+if [[ "$no_lane" == true ]]; then
+  lane=""
+fi
 
 action="${1:-list}"
 if [[ "$action" == "list" || "$action" == "login" || "$action" == "status" || "$action" == "run" ]]; then
@@ -340,6 +505,7 @@ case "$action" in
             exit 2
           }
         fi
+        capture_lane_window
         if claude_run_is_interactive "$@"; then
           start_profile_tmux "$profile" "$@"
         fi
@@ -376,8 +542,26 @@ case "$action" in
           "$mcp_sync" sync-claude-profile "$family" "$config_dir/.claude.json"
           exit 0
         fi
+        resolve_lane "$@"
+        lane_argv=(lane-start)
         if [[ -n "${lane:-}" ]]; then
-          CLAUDE_CONFIG_DIR="$config_dir" CLAUDE_BIN="$claude_bin" exec lane-start "$lane" -- \
+          if [[ "$lane_source" == record ]]; then
+            # A lane read from the record is a guess about a window nobody has
+            # named yet, so lane-start must ask before it takes it — and a
+            # lane-start that cannot ask is not handed the guess at all.
+            if lane_start_supports --confirm; then
+              lane_argv+=(--confirm)
+            else
+              echo "pclaude: $lane is the lane this workstation last paused for a swap, but this lane-start cannot ask before it takes a window (no --confirm); leaving it." >&2
+              lane=""
+              lane_note_no_lane
+            fi
+          elif lane_start_supports --yes; then
+            lane_argv+=(--yes)
+          fi
+        fi
+        if [[ -n "${lane:-}" ]]; then
+          CLAUDE_CONFIG_DIR="$config_dir" CLAUDE_BIN="$claude_bin" exec "${lane_argv[@]}" "$lane" -- \
             --allow-dangerously-skip-permissions \
             --dangerously-skip-permissions \
             --permission-mode bypassPermissions \
--- a/devcontainer.test/test-claude-profile-lane-default.sh
+++ b/devcontainer.test/test-claude-profile-lane-default.sh
@@ -1,0 +1,475 @@
+#!/usr/bin/env bash
+# Regression tests for claude-profile's LANE DEFAULT — lane-collision-protocol
+# Amendment 8(c). `--lane` used to be opt-in; a `run` that starts a
+# conversation now resolves the lane itself, in this order:
+#
+#   1. --lane / CLAUDE_LANE            -> handed to lane-start as --yes
+#   2. the current tmux window's name, when `lanes-edit.sh register-row`
+#      answers 0 for it                -> handed to lane-start as --yes
+#   3. `lanes-edit.sh swapped <ws>`, first row
+#                                      -> handed to lane-start as --confirm
+#   4. nothing                         -> today's behaviour, plus one line
+#
+# and never refuses: a lanes-edit.sh with no `swapped` subcommand (exit 2), a
+# lane-start with no `--confirm`, a missing lanes-edit.sh and `--no-lane` all
+# fall back to 4. Everything the launcher shells out to is faked here — tmux,
+# lanes-edit.sh, lane-start and claude — so the assertions are about the
+# launcher's own resolution and nothing else.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+LAUNCHER="${1:-$REPO_ROOT/base-image/files/claude-profile}"
+
+# Ambient lane/tmux state from the shell running this test — including a shell
+# that is itself a claude-profile-launched tmux child — must not reach the
+# launcher: every scenario below sets the tmux and lane state it means to test.
+unset TMUX TMUX_PANE WORKBENCHES_CLAUDE_TMUX WORKBENCHES_CLAUDE_TMUX_CHILD \
+    WORKBENCHES_CLAUDE_WINDOW WORKBENCHES_TMUX_SESSION WORKBENCHES_TMUX_PANE \
+    CLAUDE_LANE CLAUDE_NO_LANE LANES_WORKSTATION 2>/dev/null || true
+
+TEST_ROOT="$(mktemp -d)"
+trap 'rm -rf "$TEST_ROOT"' EXIT
+
+fail() {
+    echo "FAIL: $*" >&2
+    exit 1
+}
+
+PROFILE_BASE="$TEST_ROOT/profiles-home"
+PROFILE_DIR="$PROFILE_BASE/profiles/opensoft/team/team-002"
+MANIFEST="$TEST_ROOT/claude-profiles.json"
+FAKE_BIN="$TEST_ROOT/bin"
+FAKE_CLAUDE="$FAKE_BIN/claude"
+CLAUDE_LOG="$TEST_ROOT/claude.log"
+TMUX_LOG="$TEST_ROOT/tmux.log"
+LANE_START_LOG="$TEST_ROOT/lane-start.log"
+LANES_EDIT_LOG="$TEST_ROOT/lanes-edit.log"
+ERR_LOG="$TEST_ROOT/stderr.log"
+# A HOME of this test's own: lanes_edit_bin's last resort is the
+# ~/projects/xFactory/lanes-edit.sh symlink the protocol names, and the real
+# one is installed on the workstation that runs this.
+FAKE_HOME="$TEST_ROOT/home"
+mkdir -p "$PROFILE_DIR" "$FAKE_BIN" "$FAKE_HOME"
+
+printf '%s\n' \
+    '{"profiles":[{"name":"team-002","email":"test@example.invalid","family":"testing","aliases":["team002"],"profilePath":"opensoft/team/team-002"}]}' \
+    > "$MANIFEST"
+printf '%s\n' '{"name":"team-002","family":"testing","email":"test@example.invalid","aliases":["team002"]}' \
+    > "$PROFILE_DIR/.profile.json"
+
+cat > "$FAKE_CLAUDE" <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+printf '%s\n' "$*" >> "${FAKE_CLAUDE_LOG:?}"
+EOF
+
+# `display-message -p '#W'` is the window name the launcher resolves a lane
+# from; '#S' and '#{pane_id}' are export_tmux_identity's, unrelated.
+cat > "$FAKE_BIN/tmux" <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+printf '%s\n' "$*" >> "${FAKE_TMUX_LOG:?}"
+if [[ "${1:-}" == display-message ]]; then
+    case "${3:-}" in
+        '#W') printf '%s\n' "${FAKE_TMUX_WINDOW:-claude}" ;;
+        *) printf 'fake\n' ;;
+    esac
+fi
+EOF
+
+# The lane register, read-only, as Amendment 8 has the launcher read it:
+#   register-row <name>  0 = <name> is a lane with a row, 8 = no row,
+#                        2 = not even a lane-shaped name
+#   swapped [<ws>]       0 = rows `<lane>\t<UTC>\t<window>`, most recent
+#                        first; 8 = none; 2 = a lanes-edit.sh from before
+#                        Amendment 8, which has no such subcommand
+cat > "$FAKE_BIN/lanes-edit.sh" <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+{
+    printf 'argv=%s\n' "$*"
+    printf 'LANES_NO_FETCH=%s\n' "${LANES_NO_FETCH:-}"
+} >> "${FAKE_LANES_EDIT_LOG:?}"
+case "${1:-}" in
+    register-row)
+        if [[ -n "${FAKE_LANE_WITH_ROW:-}" && "${2:-}" == "${FAKE_LANE_WITH_ROW}" ]]; then
+            printf '| %s | row |\n' "$2"
+            exit 0
+        fi
+        exit "${FAKE_REGISTER_ROW_OTHER:-8}"
+        ;;
+    swapped)
+        if [[ "${FAKE_SWAPPED_STATUS:-8}" -eq 0 || -n "${FAKE_SWAPPED_ROWS_ANYWAY:-}" ]]; then
+            printf '%b' "${FAKE_SWAPPED_ROWS:-}"
+        fi
+        exit "${FAKE_SWAPPED_STATUS:-8}"
+        ;;
+esac
+echo "unknown subcommand '${1:-}'" >&2
+exit 2
+EOF
+
+# lane-start logs the argv it was launched with, and answers --help with
+# whatever help this scenario says it has — which is how the launcher decides
+# whether this lane-start knows --confirm/--yes at all.
+cat > "$FAKE_BIN/lane-start" <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+if [[ "${1:-}" == --help ]]; then
+    printf '%s\n' "${FAKE_LANE_START_HELP:-}"
+    exit 0
+fi
+printf '%s\n' "$*" >> "${FAKE_LANE_START_LOG:?}"
+EOF
+
+chmod +x "$FAKE_CLAUDE" "$FAKE_BIN/tmux" "$FAKE_BIN/lanes-edit.sh" "$FAKE_BIN/lane-start"
+
+AMENDMENT_8_HELP='OPTIONS
+  --dir <path>     the lane s checkout
+  --confirm        ask on the tty before taking a window whose name is not
+                   the lane
+  --yes            skip that question
+  -h, --help'
+OLD_HELP='OPTIONS
+  --dir <path>     the lane s checkout
+  --dry-run        print every step
+  -h, --help'
+
+# PATH carries the fakes and nothing else that could answer for them: a real
+# lane-start and a real lanes-edit.sh are installed in ~/.local/bin and
+# ~/projects/xFactory on a workstation that has run brett-wip's link-estates.
+# TMUX set means claude_run_is_interactive is false, so the launcher resolves
+# the lane in this process instead of re-exec'ing into a new tmux session
+# (scenario 10 covers that side).
+common_env=(
+    "PATH=$FAKE_BIN:/usr/bin:/bin"
+    "HOME=$FAKE_HOME"
+    "CLAUDE_BIN=$FAKE_CLAUDE"
+    "CLAUDE_PROFILES_HOME=$PROFILE_BASE"
+    "CLAUDE_PROFILES_MANIFEST=$MANIFEST"
+    "FAKE_CLAUDE_LOG=$CLAUDE_LOG"
+    "FAKE_TMUX_LOG=$TMUX_LOG"
+    "FAKE_LANE_START_LOG=$LANE_START_LOG"
+    "FAKE_LANES_EDIT_LOG=$LANES_EDIT_LOG"
+    "FAKE_LANE_START_HELP=$AMENDMENT_8_HELP"
+    "WORKBENCHES_SHARED_MCP_FAMILIES=disabled"
+    "LANES_WORKSTATION=Eagle"
+    "TMUX=fake-session"
+)
+
+claude_args='--allow-dangerously-skip-permissions --dangerously-skip-permissions --permission-mode bypassPermissions'
+note='no lane for this window; run lane-start <repo> <n> inside it'
+
+reset_logs() {
+    rm -f "$CLAUDE_LOG" "$TMUX_LOG" "$LANE_START_LOG" "$LANES_EDIT_LOG" "$ERR_LOG"
+}
+
+# launch <scenario env>... -- <launcher args>...
+launch() {
+    local -a scenario_env=()
+    while [[ $# -gt 0 && "$1" != "--" ]]; do
+        scenario_env+=("$1")
+        shift
+    done
+    shift
+    reset_logs
+    env "${common_env[@]}" "${scenario_env[@]}" "$LAUNCHER" "$@" \
+        >/dev/null 2>"$ERR_LOG"
+}
+
+lane_start_argv() { cat "$LANE_START_LOG" 2>/dev/null || true; }
+
+# ---------------------------------------------------------------------------
+# 0. --help documents the default and the way out of it.
+env "${common_env[@]}" "$LAUNCHER" --help > "$TEST_ROOT/help.out"
+grep -q -- '--no-lane' "$TEST_ROOT/help.out" || fail "--help does not document --no-lane"
+grep -q 'CLAUDE_NO_LANE' "$TEST_ROOT/help.out" || fail "--help does not document CLAUDE_NO_LANE"
+grep -q 'THE LANE DEFAULT' "$TEST_ROOT/help.out" || fail "--help does not document the lane default"
+grep -q -- '--lane' "$TEST_ROOT/help.out" || fail "--help no longer documents --lane"
+grep -q 'CLAUDE_LANE' "$TEST_ROOT/help.out" || fail "--help no longer documents CLAUDE_LANE"
+
+# ---------------------------------------------------------------------------
+# 1. The window's own name, when the register has a row for it: the lane, and
+# --yes — the operator is standing in the lane's window, there is nothing to
+# ask. The record is not even read.
+launch \
+    "FAKE_TMUX_WINDOW=openRepoProject-1" \
+    "FAKE_LANE_WITH_ROW=openRepoProject-1" \
+    "FAKE_SWAPPED_STATUS=0" \
+    "FAKE_SWAPPED_ROWS=someOtherLane-9\t2026-09-12T17:04Z\tclaude-x:0\n" \
+    -- run team002 --resume session-w
+[[ -e "$LANE_START_LOG" ]] || fail "window lane: lane-start was never invoked"
+grep -Fxq -- "--yes openRepoProject-1 -- $claude_args --resume session-w" "$LANE_START_LOG" \
+    || fail "window lane: lane-start argv was '$(lane_start_argv)'"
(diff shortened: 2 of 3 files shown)


brettheap (commented):
Lane: openRepoProject-1

**Adversarial review of the Amendment 8 launcher half.** Head **`9fdaae2`**, read-only. Text reviewed at `brettheap/new-workstation#16` `3f8df6f`; the brett-wip half at `opensoft/brett-wip#4` `fafc739`.

**Tests, run by me from the worktree, on the host:**

```console
$ bash devcontainer.test/test-claude-profile-lane-default.sh base-image/files/claude-profile
claude-profile lane default tests passed        # exit 0
$ git status --porcelain
```

Fourteen scenario blocks (0–10, plus 2b, 4b and the two `tty_launch` cases). I did not re-run the sixteen mutations; the scenarios themselves read as tight and specific, and every row of the PR's degrade table maps to one except the row named in F-W5.

**Degradation, verified against the real `opensoft/brett-wip` `main` rather than a fake:**

```console
$ ~/projects/brett-wip/lanes/lane-start --help >/dev/null 2>&1; echo $?
0
$ ~/projects/brett-wip/lanes/lane-start --help 2>&1 | grep -c -- "--confirm\|--yes"
0
$ grep -c "session-start\|swapped)" ~/projects/brett-wip/lanes/lanes-edit.sh
0
```

So today's degrade path is real: `--help` exits 0 and offers neither flag, and `lanes-edit.sh` has neither subcommand. A record-derived lane is dropped, a window lane is handed over bare — which is byte-for-byte today's `--lane` behaviour. That half is right.

### BLOCKING

**F-W1 — "Nothing in the resolution can refuse a launch" is false as the two halves compose.** `:564` `exec`s `lane-start`, so `lane-start`'s status is the launcher's. `lane-start:648` exits **2** when `--confirm` is answered `N`. An operator who says "no, not that lane" gets an exited pane and no Claude — the opposite of the invariant this PR, the amendment's clause (c) and `docs/claude-multi-account-profiles.md` all state. The **no-tty** exit 2 is clause (d)'s and is correct; the explicit-`No` 2 is not.
*My opinion on where to fix it:* in `lane-start` (F-B4 on `opensoft/brett-wip#4`) — it should print the reason and `exec` Claude bare, exit 0. If Brett rules the other way, then this file must stop using `exec` for the `--confirm` case and instead run `lane-start`, test its status, and fall through to the bare `exec "$claude_bin"` on 2. Either way it is one line, and it needs a ruling before this merges.

**F-W8 — this file is the reason the Amendment 8 hook and skill never run, and it is also where the fix is cheapest.** Every `run` execs with `CLAUDE_CONFIG_DIR="$config_dir"` (`:524`, `:564`, `:570`), so the harness reads the **profile's** `settings.json` and the profile's `skills/`, never `~/.claude/`. Measured on this workstation: the live profile's `settings.json` has `hooks` = `{UserPromptSubmit: …}` only, the profile's `skills` is a symlink to `/home/brett/.claude-profiles/shared/skills`, and that directory is **empty**. The amendment installed both pieces into `~/.claude/`, where nothing under this launcher reads them.
The good news is in this file: the jq at `:55-62` does `.hooks = (.hooks // {}) | .hooks.UserPromptSubmit = [ … ]` — it **preserves** every other key under `hooks`. So a `SessionStart` entry written into a profile's `settings.json` survives the per-launch rewrite untouched, and no change to this file is needed for it. *Recommendation for this PR:* add one sentence to `docs/claude-multi-account-profiles.md` saying that the per-launch settings rewrite is additive for hook kinds other than `UserPromptSubmit`, so an installer may write `hooks.SessionStart` into a profile's settings and rely on it surviving. Amendment 9's `openRepoTools --install` then has a documented seam. Full evidence is in the #15 summary under *Lead 2* (F-S1/F-S2).

### minor

- **F-W2 — the `--yes` probe buys nothing.** `:559` `elif lane_start_supports --yes; then lane_argv+=(--yes)`. `--yes` is inert in `lane-start` unless `--confirm` was also passed (`lane-start:612` gates the whole block on `confirm`), so on cases 1 and 2 the flag changes nothing and its absence changes nothing. Either drop the probe and the flag for those cases, or pass `--yes` only alongside `--confirm` — which is the only combination that means anything.
- **F-W3 — two stderr lines for one situation.** The `--confirm` degrade prints its own "…cannot ask before it takes a window…; leaving it." and then `lane_note_no_lane` (`:557`). One operator, one situation, two lines that say overlapping things. Make it one, and let it name the lane it declined.
- **F-W4 — `swapped` exit 2 is read as "this `lanes-edit.sh` predates Amendment 8", but 2 is also its usage code** (`lanes-edit.sh:3191`, `:3192`). A caller bug degrades silently to case 4 and looks exactly like an old helper. Probe `swapped` the way `lane-start` is probed, or ask brett-wip for a distinct code (F-B8).
- **F-W5 — one degrade-table row has no scenario.** "no `lane-start` on `PATH` — no lane estate at all → silent; not even the note" is implemented at `:361` (`command -v lane-start … || return 0`) but no scenario in `test-claude-profile-lane-default.sh` covers it; scenario 9 covers the missing `lanes-edit.sh`, which is a different branch and a different outcome (the note *is* printed). Add it — a mutation that turned that `return 0` into a `lane_note_no_lane` would pass the suite today.
- **F-W6 — the test prints one line and no count.** `claude-profile lane default tests passed` is all a reader gets. A scenario silently skipped, or a `launch` that never ran, is invisible. Print `N scenarios, M assertions` the way the brett-wip suite does — it is what let me confirm that side's 760 in one line.
- **F-W7 — the docs repeat the false invariant.** `docs/claude-multi-account-profiles.md`: "Nothing in the resolution can refuse a launch: a `lanes-edit.sh` with no `swapped` subcommand, or a `lane-start` with no `--confirm`, falls back to the unchanged behaviour." The two examples given are true; the sentence they are attached to is not, for the reason in F-W1.
- **F-W9 — the LANES_NO_FETCH reads are right, and are also the amendment's strongest unforced argument.** Both register reads carry `LANES_NO_FETCH=1` (`:376`, `:386`) and the test asserts it. Worth saying explicitly in the PR body that this is what makes the default safe to put on the path of every launch: `register-row` and `swapped` both still read `origin/main`, just as it already stands here. The counterpart problem — a read that does *not* set it, in front of every session start — is F-B2 on `opensoft/brett-wip#4`.

### What I checked and found sound
The resolution order and its precedence; `--no-lane`/`CLAUDE_NO_LANE` winning over `--lane` in either order (`:460` clears `lane` after the option loop, and `resolve_lane`'s early return); the `claude_args_start_session` split out of `claude_run_is_interactive`, so `mcp`/`doctor`/`--print` neither re-exec nor resolve; `lanes_edit_bin`'s four-step search with no workstation-specific path; `lane_start_help`'s `${var+set}` caching idiom (correct — it caches an empty answer too); the tmux-child guard, which is the half of the carry that actually does work, and the scenario that pins a re-exec carrying no window name. The PR body's honesty about the unreachable carry and about `register-row` fetching is the right call and made this review much faster.

### Opinion
**NOT READY**, on F-W1 alone — a launcher that can strand an operator in an exited pane is the one failure this change exists to prevent, and it needs Brett's ruling on which half moves. F-W8 does not block *this* PR's code but does block the amendment's adoption act, and one documentation sentence here would close it. Everything else is minor. This is a review opinion, not a ratification.

brettheap:
Lane: openRepoProject-1

**Fix round, progress 1 of 2 — pushed `fa00129`.** R-A8-5 is in: `claude-profile` ensures clause (e)'s `SessionStart` entry in the *profile's* `settings.json` on every run (exact-string match on the command, additive, `timeout: 5`), and `setup-claude-profiles.sh` installs the vendored `/lane-swap` skill into `~/.claude-profiles/shared/skills/` — the one directory all 418 profiles' `skills` symlink already points at.

Installed by hand on Eagle at the same time (the launcher does the 418 per run once this lands): the skill at `/home/brett/.claude-profiles/shared/skills/lane-swap/SKILL.md` and `~/.claude/skills/lane-swap/SKILL.md`, and `~/.claude/settings.json`'s entry replaced with the snippet below (backup `settings.json.bak-20260912T193345Z`; everything outside `hooks.SessionStart` verified byte-identical).

**Measured proof the R-A8-5(a) diagnosis was right:** the moment the file landed in the shared directory, this session's own available-skills list gained `lane-swap` — under `CLAUDE_CONFIG_DIR=…/profiles/opensoft/team/team-05f`, which is exactly where F-S1 said nothing was being read from.

Still to come in round 2: R-A8-3 (bare launch on a declined `--confirm`), F-W2/F-W3/F-W4, the new scenarios and the mutation set, and the docs.

brettheap:
Lane: openRepoProject-1

**Fix round, progress 2 — pushed `9cfd67c`.** R-A8-3's launcher half and the F-W minors.

The `--confirm` path no longer `exec`s: it runs `lane-start`, and an exit 2 falls through to a bare Claude with one notice, exit 0. A 2 from a `lane-start` that *did* take the lane is handed back instead — the window name says which happened, because taking a lane renames it and declining leaves it alone, so the launcher can never start a second Claude behind the first. `--yes` is passed nowhere (F-W2). One stderr line per situation, naming the lane (F-W3). A `swapped` that exits 2 *with* the subcommand present is reported as a caller bug rather than read as an old helper (F-W4). The no-estate row has its scenario (F-W5). Two more from Copilot's suppressed comments: a one-field untabbed `swapped` row is no longer read as a lane (the tab is the contract), and the record is not read at all outside tmux, where `lane-start` has no window to rename.

Also clause (e)'s rule that the hook's block is the only no-lane notice in the estate: the launcher ensures that hook, so it knows when it will speak, and stands its own note down when it will — and keeps printing it where the estate has no `session-start` subcommand yet.

`devcontainer.test/test-claude-profile-lane-default.sh`: **30 scenarios, 101 assertions**, up from 25 and an unprinted zero (F-W6).

Next in this round: the hook entry moved to clause (e)'s exact string with a capability probe, the `setup-claude-profiles.sh` ruling from workBenches#68 F5, the skill's remaining fixes, the two new suites, the mutation set and the docs.

brettheap:
Lane: openRepoProject-1

**Fix round, progress 3 — pushed `15bf212`.** R-A8-5's two halves, their suites, and the docs.

**The entry is the canonical string, byte for byte** (ruled this round; `|| true` stays):

```
~/projects/xFactory/lanes-edit.sh session-start || true
```

with `"matcher": "startup|resume|clear|fork"` and `"timeout": 5`. I had briefly removed `|| true` to match the #16 snippet as it read at `origin/lane-protocol/amendment-8-swap-restart` — that text says the command is bare "since the subcommand exits 0 by construction" — and reverted on the ruling. **The difference the text still has to follow is exactly `|| true`.**

One thing the ruling does not cover, so it is stated rather than assumed: `|| true` hides the *status* of a `lanes-edit.sh` with no `session-start`, not the usage block it prints, and nothing may add `2>/dev/null`. So the ensure is **gated on the subcommand existing** as well as the script — one local grep, no execution, no network — and the entry appears on the first run after brett-wip#4 lands, because the ensure runs on every run. Without that gate, every session start on this workstation would carry a usage block until that PR merges. One line to drop if you want it written unconditionally.

The setup loop that installs the skill is now idempotent **by content** and carries the workBenches#68 F5 sentence: it stands for A8, is the sole writer of those paths until A9's act 3, and A9's act 4 deletes it.

**Suites** (all three green on the host, this head):

| suite | result |
|---|---|
| `test-claude-profile-lane-default.sh` | 30 scenarios, 101 assertions |
| `test-claude-profile-session-start-hook.sh` (new) | 10 scenarios, 25 assertions |
| `test-claude-profile-skill-install.sh` (new) | 4 scenarios, 33 assertions |

The two container-capable ones are registered in `devcontainer.test/test.sh`, which Copilot asked for twice; the skill one is host-run like `test-setup-estate-commands.sh` and the README says so.

Still to come: the mutation set, and the hand-installed copies on this workstation brought to the canonical string.

brettheap:
Lane: openRepoProject-1

**Fix round complete — head `8eb1343`.** Every finding on [review 5187711401](https://github.com/opensoft/workBenches/pull/63#pullrequestreview-5187711401), the skill and hook findings F-S1–F-S9 from the [#15 summary](https://github.com/brettheap/new-workstation/issues/15#issuecomment-5648122921), and the rulings of [A8 Addendum 2](https://github.com/brettheap/new-workstation/issues/15#issuecomment-5648139625). Three commits since the review, plus `fa00129` from round 1.

## Finding → commit

| finding | what changed | commit |
|---|---|---|
| **F-W1** BLOCKING — an `N` to `--confirm` left an exited pane | the confirm path RUNS `lane-start` instead of `exec`ing it and falls through to a bare Claude on 2, exit 0, one notice | `9cfd67c` |
| **F-W2** — the `--yes` probe buys nothing | `--yes` is passed nowhere; cases 1 and 2 go over bare | `9cfd67c` |
| **F-W3** — two stderr lines for one situation | one line, and it names the lane it left | `9cfd67c` |
| **F-W4** — `swapped` exit 2 is also the usage code | the helper's own words tell them apart: an old helper is silent, a caller bug gets one line, both still launch | `9cfd67c`, scenario fixed in `8eb1343` |
| **F-W5** — the no-estate row had no scenario | scenario 9a, the one silent row | `9cfd67c` |
| **F-W6** — the suite printed no count | `31 scenarios, 106 assertions`, scenario count pinned | `9cfd67c` |
| **F-W7** — the docs repeat the false invariant | the docs now say what makes it true | `15bf212` |
| **F-W8** — this file is why the hook and skill never ran | the launcher ensures the hook in the profile's settings; the docs state the additivity an installer may rely on | `fa00129`, `15bf212` |
| **F-W9** — say `LANES_NO_FETCH` is what makes it safe | in the PR body | body |
| **F-S1/F-S2** — skill and hook installed where nothing reads them | shared skills dir; profile settings | `fa00129`, `15bf212` |
| **F-S3** — the installed entry was unbounded | matcher + `"timeout": 5`, the canonical string | `fa00129`, `15bf212` |
| **F-S4** — the hook calls a subcommand that does not exist yet | the entry is not written until `lanes-edit.sh` HAS `session-start` | `15bf212` |
| **F-S5–F-S9** — the skill's name, the ` — ` sanitise, the derived state word, the capability probe, ONE restart command | the rewritten skill, now pinned by a suite | `fa00129`, `15bf212` |
| Copilot: `cut -f 1` takes a whole untabbed line | the tab is the contract | `9cfd67c` |
| Copilot: a record lane outside tmux goes to a `lane-start` that cannot take it | the record is not read outside tmux | `9cfd67c` |
| Copilot: the suite is not in the Layer 0 runner | both container-capable suites registered in `test.sh` | `15bf212` |
| Copilot: the skill's literal `<repo> <n>` | filled in from the window name split on its last `-` | `15bf212` |
| Copilot: the skill ignores `swapped`'s exit status | it reads the record exactly as the launcher does | `15bf212` |
| Copilot: derive the hook's command from the resolved `lanes-edit.sh` | **not done, by design** — the exact command string is the idempotence key, so it must be identical on every machine and in every profile, and clause (e) allows a host path in a settings file precisely because a repo-relative one cannot be one. The existence check names the same path the hook will call. | — |

## Ruling → commit

| ruling | commit |
|---|---|
| **R-A8-3** — a launcher never refuses to start Claude; both answers lead somewhere | `9cfd67c` |
| **R-A8-5(a)** — the skill into the shared skills directory, idempotent | `fa00129`, `15bf212` |
| **R-A8-5(b)** — the launcher ensures the hook on every run, idempotent by exact-string match, bounded | `fa00129`, `15bf212` |
| **R-A8-6** — `/resume` is not a lane surface: the skill offers no such door | `fa00129`, pinned `15bf212` |
| **R-A8-7** — one no-lane notice, not two; the F-W minors; the skill's F-S items | `9cfd67c`, `15bf212` |
| workBenches#68 **F5** — the setup loop stands for A8, sole writer until A9's act 3, deleted by act 4; idempotent by content | `15bf212` |
| the canonical hook string (this round's ruling) | restored in `15bf212` |

## Counts

| suite | scenarios | assertions |
|---|---|---|
| `test-claude-profile-lane-default.sh` | 31 | 106 |
| `test-claude-profile-session-start-hook.sh` (new) | 10 | 25 |
| `test-claude-profile-skill-install.sh` (new) | 4 | 34 |
| **total** | **45** | **165** |

**Mutations: 39 run, 39 caught.** Three escaped the first pass and each was a real hole — a `swapped` row of the right shape whose first field is not a lane name had no scenario; `grep -c` counts lines, so a `claude --resume` fallback could hide on the prohibition's own line; and **scenario 4 had never tested what it claimed**: the fake `lanes-edit.sh` selected its old-helper branch with `break` inside a `case`, which is meaningless outside a loop, so bash warned and ran the `swapped` arm anyway. The F-W4 code path had never been executed by the test written for it. All three closed in `8eb1343`.

Also green on this head, on the host: `test-claude-profile-lane-start.sh`, `test-claude-profile-preserves-model.sh`, `test-claude-tmux-statusline.sh`, `test-claude-tmux-statusline-env-isolation.sh`, `test-claude-statusline-snapshots.sh`, and `bash -n` on all six changed shell files. No shellcheck on this host this round.

## The exact hook string matched

```
~/projects/xFactory/lanes-edit.sh session-start || true
```

with `"matcher": "startup|resume|clear|fork"` and `"timeout": 5`. Idempotence matches on that command string and on nothing else.

**Difference from #16 as I read it:** at `origin/lane-protocol/amendment-8-swap-restart` clause (e) gives the command **bare** and says in terms that it carries "no `|| true`, since the subcommand exits 0 by construction". I briefly matched that text, and reverted on the ruling that the string above is canonical and the text follows it. **The one difference the text has to absorb is exactly `|| true`.** The rest — matcher, `timeout: 5`, no `bash -lc`, no `2>/dev/null`, exact-string idempotence, "ensured on every run, not installed once" — already agrees byte for byte.

One thing the ruling does not cover, stated rather than assumed: `|| true` hides the *status* of a `lanes-edit.sh` with no `session-start`, not the usage block it prints, and nothing may add `2>/dev/null`. So the write is also gated on the subcommand existing — a local grep of one regular file, no execution, no network — and the entry appears on the first run after brett-wip#4 lands. Without that gate every session start on this workstation carries a usage block until that PR merges. It is one line to drop if you want it written unconditionally.

## Installed by hand on this workstation (Eagle)

| path | state | backup |
|---|---|---|
| `/home/brett/.claude-profiles/shared/skills/lane-swap/SKILL.md` | refreshed to this head's vendored text | `SKILL.md.bak-20260912T224350Z` |
| `/home/brett/.claude/skills/lane-swap/SKILL.md` | refreshed to this head's vendored text | `SKILL.md.bak-20260912T224350Z` |
| `~/.claude/settings.json` | **unchanged this round** — verified already carrying the canonical entry byte for byte | round 1's `settings.json.bak-20260912T193345Z` |

No profile `settings.json` was hand-edited, and none needed to be: re-measured just now, **418 profile `settings.json` files, 0 with a `SessionStart` hook**, and today's `lanes-edit.sh` has no `session-start`, so the launcher would write nothing yet either. `~/.local/bin/pclaude` still resolves to the live `workBenches` checkout, which is not this branch — nothing here is live until merge.

Head: **`8eb1343`**. Still a draft; not marked ready, not merged.

brettheap (commented):
Lane: openRepoProject-1

**Independent RE-VERIFICATION of the Amendment 8 launcher half, head `8eb1343`.** Read-only: no fixes, no commits, nothing marked ready. All three suites and all three mutations run by me from the worktree, restored exactly; `git status --porcelain` empty and `HEAD` at `8eb1343` after each.

### Suites — the round-up's counts confirmed exactly

```console
$ bash devcontainer.test/test-claude-profile-lane-default.sh base-image/files/claude-profile
claude-profile lane default: 31 scenarios, 106 assertions passed          # exit 0
$ bash devcontainer.test/test-claude-profile-session-start-hook.sh base-image/files/claude-profile
claude-profile SessionStart hook: 10 scenarios, 25 assertions passed      # exit 0
$ bash devcontainer.test/test-claude-profile-skill-install.sh
lane-swap skill install: 4 scenarios, 34 assertions passed                # exit 0
```

**45 scenarios, 165 assertions.** F-W6 is properly closed: a skipped scenario is now visible.

### Mutations — three applied, three caught, three restored

| mutation | result |
|---|---|
| drop `\|\| true` from `lane_session_start_command` (`:34`) | **caught** — hook suite fails: *"ensure: matcher was , expected `startup\|resume\|clear\|fork`"* |
| `exec` `lane-start` again on the record path (`:733-737`) — the F-W1 regression | **caught** — lane-default fails: *"declined confirm: Claude was not launched bare"* |
| `session_start_ok=true` unconditionally, dropping the capability probe (`:101`) | **caught twice** — hook suite: *"no subcommand: an entry was written against a lanes-edit.sh that has none"*; lane-default: *"untabbed record row: the note was not printed"* |

The first is the one I most wanted to see bite, because the entry's idempotence is a whole-string match and a one-byte drift is a duplicate block at every session start for ever.

### The hook string, measured across all four surfaces

`:34` / `:35` / `:36` are byte-identical to `brettheap/new-workstation#16` @`ab44842` lines **523**/**530**/**532**, to Amendment 9's `#18` @`2b912bc` lines **341**/**343**, and to the entry installed in `/home/brett/.claude/settings.json` on Eagle:

```
command  ~/projects/xFactory/lanes-edit.sh session-start || true
matcher  startup|resume|clear|fork
timeout  5
```

The ensure jq (`:48-58`) appends only when no entry's `.command` equals that string and touches nothing else under `.hooks` — idempotent exactly as clause (e) specifies.

### Findings review-by-review

`F-W1` closed (mutation 2; `:733-747` runs `lane-start`, tests the status, and starts Claude bare on a 2 that left the window unrenamed). `F-W2` closed (`--yes` passed nowhere, `:711-714`). `F-W3` closed (`lane_note_no_lane` takes the caller's reason and prints **one** line, `:419-443`). `F-W4` closed (the helper's own words tell an old helper from a caller bug, `:540-544`). `F-W5` closed (scenario 9a). `F-W6` closed. `F-W7` closed (`docs/claude-multi-account-profiles.md:187`). `F-W8` closed (`:207-212` states the additivity an installer may rely on). `F-W9` — both register reads carry `LANES_NO_FETCH=1` (`:495`, `:521`). `F-S1`–`F-S9` — see the summary on `brettheap/new-workstation#15`; the vendored skill, the shared copy and the `~/.claude` copy are byte-identical (`md5 9c6c191857d457d9c1398bd6f2f21fad` on all three).

### minor

**RV-W1 — the no-lane notice stands down on a probe of the *helper*, not on the entry being in the profile it is launching.** `lane_session_start_ensured` is set at `:106` from the capability probe at `:101-105`, before `configure_profile_runtime` has written anything, and `lane_note_no_lane` returns silently on it (`:433-436`). The two come apart in one real window: once `opensoft/brett-wip#4` lands, the probe answers **true** for every profile, but a profile that has not yet been through one `configure_profile_runtime` still has no `SessionStart` entry — so its **first** launch prints neither the launcher's notice nor the hook's block. One launch, self-healing on the next, and one line to close. *Fix:* set `lane_session_start_ensured=true` after the `mv` of `$settings_tmp` succeeds for **this** profile, not at probe time.

**RV-W2 — the `~/.claude` copy of the skill is installed and its companion hook entry is not, and nothing owns it.** `scripts/setup-claude-profiles.sh` contains **0** occurrences of `SessionStart`: its loop (`:192-202`) writes `SKILL.md` to both `$base/shared/skills/lane-swap/` and `$default_claude_dir/skills/lane-swap/`, idempotent by content — correct, and matching R-A8-5(a) and `#68` F5. But `claude-profile` ensures the entry only in the profile it is launching, and Amendment 8 act 6 says "the hook entry needs no install act of its own" on the strength of that. So a **bare `claude`** run — the one thing act 6 keeps `~/.claude` for — gets the skill and no hook. On Eagle the entry exists in `~/.claude/settings.json` only because round 1 hand-wrote it; on any other workstation it would not. *Fix:* have the setup script ensure the same entry in `$default_claude_dir/settings.json` under the same exact-string jq (it already rewrites that file a few lines below), or have #16 say in terms that the bare-`claude` hook is installed by nothing and is the operator's.

**RV-W3 — live state on Eagle, so the adoption act is not read as done.** Re-measured now: **323** profile `settings.json` under `~/.claude-profiles/profiles/*/*/*/`, **0** carrying a `SessionStart` hook; `~/.local/bin/pclaude` still resolves to `/home/brett/projects/workBenches/base-image/files/claude-profile` (the live checkout, not this branch); and `~/projects/xFactory/lanes-edit.sh` → `brett-wip/lanes/lanes-edit.sh` on `main`, which has **0** occurrences of `session-start`. So the probe correctly writes nothing today, and **the hook has still never fired under a profile launch** — F-S2's condition is fixed in code and not yet in effect, exactly as the round-up says. Worth one line in the PR body so a later reader does not take the "installed by hand" table as evidence the seam is live. (Not a code defect.)

**Observation, not a finding.** `lane_session_start_script` (`:33`) is a hard-coded `$HOME/projects/xFactory/lanes-edit.sh` while `lanes_edit_bin` has a four-step search. A workstation whose estate is not at that path would probe a file that does not exist and never ensure the entry, even with a good `lanes-edit.sh` elsewhere. Copilot raised the neighbouring point and the answer given — the existence check must name the same path the hook will call, because the command string is the idempotence key — is right, and I would not change it. It is a documented limit of pinning a host path, and it belongs in the PR body rather than in the code.

### Opinion

**READY**, as a review opinion and not a ratification. Every BLOCKING finding on `pullrequestreview-5187711401` is closed, the three suites are real and the three mutations I chose all bit in the assertion written for them, the canonical hook string agrees byte-for-byte across #16, #18 and the installed entry, and the one thing the original review said did not block this PR's code — F-W8 — is closed with the documentation seam A9 needs. RV-W1 and RV-W2 are both one-line fixes and neither can strand an operator; I would take them in this PR rather than after it, but I would not hold the PR for them. **I am not marking this ready and this is not a ratification.**

brettheap:
Lane: openRepoProject-1

**Fix round closing the re-verification findings on this PR (RV-W1–RV-W3, `pullrequestreview-5188764062`) and on the vendored `/lane-swap` skill (RV-S1–RV-S4, `brettheap/new-workstation#15` comment `5649690877`).** Read-write this time — commits below, pushed as each finding group closed. Not marked ready; not merged.

### Finding → commit

| finding | severity | commit | what changed |
|---|---|---|---|
| RV-S1 | BLOCKING | `796cc22` | Step 2's handoff derivation switched from `awk -F'|' '{print $(NF-2)}'` (right split, breaks the moment any earlier/later cell carries a literal `\|` — this lane's own row has 27/28 fields) to a fixed LEFT field index, `awk -F'|' '{print $7}'` (lanes-edit.sh's own `row_cell`/`handoff_of_lane` convention: column 6, field 7, `$1` being the empty string before the leading pipe). Verified against the real register row (`git -C ~/projects/brett-wip fetch && git show origin/main:lanes/LANES.md \| grep '^\| \`openRepoProject-1\`'` — NF=28, field 7 = `handoffs/openRepoProject/session-handoff-2026-09-11-lane-openRepoProject-1.md`, the file that's actually on disk). Added a stop-and-report guard for an empty handoff. |
| RV-S3 | minor | `796cc22` | Dropped `LANES_NO_FETCH=1` from step 2's `register-row` read only (it decides state — handoff + last session uuid — not merely which lane a window is); step 1's identity probe keeps it. |
| RV-S2 | minor | `796cc22`, `b84f26b` | Step 4 now sets `row_write_refused` when `replace-in-row`/`append-row-status` fail; step 5 reads it and prints `pclaude --lane <lane> run <profile>` (leading option, matching `claude-profile`'s actual argv order — confirmed against `#16`'s just-landed text fix, `c769cc2`) only in that case, never `pclaude run <profile> --lane <lane>` (which `claude-profile` would hand to Claude itself). |
| RV-S4 | minor, unchanged | none | Reviewer's own recommendation is "no action" — the installed `~/.claude` entry names a subcommand `lanes-edit.sh` on `brett-wip` `main` still doesn't have; fixed in code, not yet in effect until the brett-wip half lands. Re-confirmed still true (see live state below). |
| RV-W1 | minor | `8708154` | `lane_session_start_ensured` is now set after `configure_profile_runtime`'s `mv` of `$settings_tmp` succeeds for *this* profile, not at probe time — it also now correctly stays `false` if the function returns early (invalid settings JSON). |
| RV-W2 | minor | `52b80da`, `9cc7b19` | **Course-corrected mid-task per the coordinator**, after `new-workstation#16`'s text round landed (`c769cc2`) making adoption act 6 own this: `scripts/setup-claude-profiles.sh` now *installs* the identical canonical `SessionStart` entry (same command/matcher/timeout, same exact-string idempotence, additive beside other hook kinds) into `$default_claude_dir/settings.json`, gated on the same estate probe — not only documented as absent. `claude-profile`'s own comment cross-references the second writer. |
| RV-W3 | minor | `9cc7b19` + PR body | Added a "live state" note to `docs/claude-multi-account-profiles.md` and to this PR's body (the reviewer's explicit ask) so the hook-string/adoption tables are never read as a live measurement. |

### Suites — all green

```
claude-profile lane default: 31 scenarios, 106 assertions passed
claude-profile SessionStart hook: 10 scenarios, 25 assertions passed
lane-swap skill install: 9 scenarios, 52 assertions passed   # was 4 scenarios / 34 assertions
```
**50 scenarios, 183 assertions.** `test-claude-profile-lane-start.sh` and `test-claude-profile-preserves-model.sh` (untouched) still pass.

New coverage in `test-claude-profile-skill-install.sh` (scenarios 5–9, `EXPECTED_SCENARIOS` 4 → 9): the bare-`claude` hook's shape/idempotence/additivity/gating (RV-W2), and — executing the shipped SKILL.md lines themselves against fixture rows rather than a hand-copied stand-in — RV-S1's handoff derivation on a plain row and a row shaped like this lane's own (extra `\|` characters and all), and RV-S2's restart-command branch both ways.

### Mutations — 6 applied, 6 caught, all restored

Re-ran the original reviewer's 3 (no regression), then 3 new ones for this round's own fixes:

| mutation | result |
|---|---|
| drop `\|\| true` from `lane_session_start_command` (original) | **caught** — hook suite: matcher mismatch |
| `exec` `lane-start` again on the record path (original, F-W1) | **caught** — lane-default: "declined confirm: Claude was not launched bare" |
| `session_start_ok=true` unconditionally (original) | **caught twice** — hook suite + lane-default, same messages the reviewer reported |
| revert handoff derivation to `$(NF-2)` (RV-S1) | **caught** — skill-install: "the handoff derivation itself still counts back from NF" |
| remove the `row_write_refused` branch (RV-S2) | **caught** — skill-install: "step 4's row write has no refused-write branch" |
| remove the hook install from `setup-claude-profiles.sh` (RV-W2) | **caught** — skill-install: "default hook: 0 SessionStart entries, expected 1" |

`git status --porcelain` empty and `HEAD` at each commit's own sha after every restore.

### The skill, reinstalled by hand on Eagle

`/home/brett/.claude-profiles/shared/skills/lane-swap/SKILL.md` and `~/.claude/skills/lane-swap/SKILL.md`, both backed up first (`SKILL.md.bak-20260913T024211Z` beside each), then overwritten with the fixed vendored copy, mode 644.

**md5 `a28bbeb9bbf31aa34553d127a3f18430`** — identical across the vendored file, the shared install, and the `~/.claude` install.

### Live state (unchanged by this round, re-confirmed just now)

418 profile `settings.json` under `~/.claude-profiles/profiles/*/*/*/`, 0 with a `SessionStart` hook; `~/.local/bin/pclaude` → the live `workBenches` checkout, not this branch; `~/projects/xFactory/lanes-edit.sh` → `brett-wip/lanes/lanes-edit.sh` on `main`, 0 occurrences of `session-start`. Both writers of the entry (this launcher, and `scripts/setup-claude-profiles.sh` as of this round) therefore correctly ensure nothing yet — RV-S4's "no action" stands, and this round's PR-body note says so in terms.

**Head `b84f26b70ff6a6dae933f2b9a61d0d2f5282d6b2`.**

brettheap (commented):
Lane: openRepoProject-1

**CONFIRMATION pass over the Amendment 8 launcher and the vendored `/lane-swap` skill after the fix round, head `b84f26b70ff6a6dae933f2b9a61d0d2f5282d6b2`.** Read-only: no fixes, no commits, nothing marked ready. All three suites and both mutations run by me from the worktree; `git status --porcelain` **empty** and `HEAD` unmoved at `b84f26b` before and after each.

### Suites — the fix round's counts confirmed exactly

```console
$ bash devcontainer.test/test-claude-profile-lane-default.sh base-image/files/claude-profile
claude-profile lane default: 31 scenarios, 106 assertions passed          # exit 0
$ bash devcontainer.test/test-claude-profile-session-start-hook.sh base-image/files/claude-profile
claude-profile SessionStart hook: 10 scenarios, 25 assertions passed      # exit 0
$ bash devcontainer.test/test-claude-profile-skill-install.sh
lane-swap skill install: 9 scenarios, 52 assertions passed                # exit 0
```

**50 scenarios, 183 assertions.** 45/165 → 50/183, as claimed. `gh pr checks 63` → all four CodeQL checks **pass**.

### Closure table

| finding | state | evidence |
|---|---|---|
| **RV-S1** (was BLOCKING) | **closed** — see the ruling judgement below | `SKILL.md:77` `awk -F'\|' '{print $7}'`. Verified against this lane's **real published row**, read read-only after a fetch: `git -C ~/projects/brett-wip fetch && git show origin/main:lanes/LANES.md \| grep '^\| \`openRepoProject-1\`'` → **NF=28**, and field **7** is ` handoffs/openRepoProject/session-handoff-2026-09-11-lane-openRepoProject-1.md ` — the file that is on disk and the one I read this lane's *Lessons* out of. `$(NF-2)` is field 26, inside the state history. A stop-and-report guard on an empty `$handoff` is at `:83-85`. |
| **RV-S2** | **closed** | `SKILL.md:133-137` sets `row_write_refused` on a failed `replace-in-row`/`append-row-status`; `:152-153` prints `pclaude --lane $lane run ${CLAUDE_PROFILE_NAME:-<profile>}` only in that case, and `:167-171` states the leading-option rule. Matches `new-workstation#16` `c769cc2` **162-171**. |
| **RV-S3** | **closed** | `LANES_NO_FETCH=1` survives only on step 1's identity probe (`:28`); step 2's `register-row` (`:71`) is a fetched read, with `:66-68` giving the reason. |
| **RV-S4** | **stands, no action, correctly** | Re-measured on Eagle: `~/projects/xFactory/lanes-edit.sh` → `brett-wip/lanes/lanes-edit.sh` on `main`, **0** occurrences of `session-start`; **418** profile `settings.json`, **0** with a `SessionStart` hook; `~/.local/bin/pclaude` → `/home/brett/projects/workBenches/base-image/files/claude-profile`, the live checkout, not this branch. Both writers of the entry correctly ensure nothing yet. |
| **RV-W1** | **closed** | `lane_session_start_ensured=false` at `:49` with the comment stating it is set by `configure_profile_runtime` for *this* profile, not at probe time. |
| **RV-W2** | **closed** | `scripts/setup-claude-profiles.sh` went from **0** `SessionStart` occurrences to **3**: `:214-216` carry the three canonical values, `:219` gates on the same estate probe (`grep -qF 'session-start'` of the installed helper), `:228-229` is the same exact-string ensure jq. `claude-profile:34-39` cross-references the second writer and says to change both together. This is the course-correction `#16`'s act 6 asked for and it matches act 6 **856-859**. |
| **RV-W3** | **closed** | Live-state note in `docs/claude-multi-account-profiles.md` and the PR body (`9fdaae2`… `9cc7b19`), so the hook-string and adoption tables are not read as a live measurement. Re-measured above; still true. |

### RV-S1 — the ruling judgement asked for

**The ruling:** *"the skill never parses a row itself, it uses the helper's own reads."* **The fix:** a fixed **left** field, `$7`, of `register-row`'s output, replacing the right-split `$(NF-2)`.

**It satisfies the ruling, and I would not hold the PR for the letter of it.** Four reasons, in order of weight:

1. **`$7` is not a re-derivation — it is the helper's own read, character for character.** `lanes-edit.sh:1978-1982`: `handoff_of_lane() { … row_cell "$hl_row" 7 … }`, and `row_cell` at `:1957` is literally `awk -F'|' -v i="$2" '{print $i}'`. The skill runs the same awk, on the same field index, over the same row. Its `uuid` line (`:78-79`) mirrors `session_ids_of_lane` (`row_cell "$si_row" 3 \| uuids_in_cell`) plus `last_session_id_of_lane`'s `tail -n1` the same way. There is no second convention to drift from, which is what the ruling was protecting.
2. **The hazard the ruling names is structurally gone.** `lanes-edit.sh:395-399` gives the reason a right-split cannot work — *"the only safe way to reach cell 3 of a row whose LATER cells may contain a literal `|` … It splits from the LEFT for that reason, never from NF"* — and `:1972-1977` says it again for the handoff column specifically. A **left** index is immune to everything to its right by construction, and on this lane's row that is all 20 extra pipes: field 7 reads correctly at NF=28, which is itself the proof that the extras are in the state cell.
3. **The literal reading is not reachable from the skill.** `handoff_of_lane` is an internal shell function with exactly two callers (`:2295` in `who`, `:2903` in `session_start_block`) and **no subcommand exposes it** — I checked the whole dispatch. "Use the helper's own reads" with no read to call would mean adding a subcommand to `opensoft/brett-wip#4`, which is a different PR and a different round. Taking the helper's convention was the available way to obey the ruling, and it is the one taken.
4. **The residual is exactly the helper's residual, not a new one.** A left index is exposed to a literal `|` in a cell to its *left* — lane, session, workstation, started, objects. `handoff_of_lane` carries that identical exposure today. The skill is no less safe than the code it mirrors, and `:83-85` now turns the failure mode from a silent `git add ""` into a stop-and-report, which is the harm the finding was actually about.

**Verified on the live row and not blocking.** The field is right on the one row the finding was measured broken on.

### Mutations — two spot-checks, both caught, both restored

| mutation | result |
|---|---|
| revert the handoff derivation to `$(NF-2)` (`SKILL.md:77`) — this round's own fix | **caught** — skill-install, exit 1: *"FAIL: RV-S1: the handoff derivation itself still counts back from NF: `handoff="$(printf '%s' "$row" \| awk -F'\|' '{print $(NF-2)}' \| awk '{$1=$1};1')"`"* — and it quotes the offending line back, so the failure names the defect rather than a count |
| `exec` `lane-start` again on the record path (`claude-profile:750-754`) — the **F-W1 regression check** | **caught** — lane-default, exit 1: *"FAIL: declined confirm: Claude was not launched bare ()"* |

Each restored by `cp` from a pre-mutation byte-identical copy; `git status --porcelain` empty and `HEAD` at `b84f26b` after both. The second is the one I most wanted to re-run: `RV-W1`'s and `RV-W2`'s commits both touch the same region of `claude-profile`, and the guard they sit beside is the whole of R-A8-3.

### Skill + hook as installed on Eagle

| file | md5 |
|---|---|
| `base-image/files/claude/skills/lane-swap/SKILL.md` (vendored, `b84f26b`) | `a28bbeb9bbf31aa34553d127a3f18430` |
| `/home/brett/.claude-profiles/shared/skills/lane-swap/SKILL.md` | `a28bbeb9bbf31aa34553d127a3f18430` |
| `/home/brett/.claude/skills/lane-swap/SKILL.md` | `a28bbeb9bbf31aa34553d127a3f18430` |

Byte-identical across all three, and equal to the figure in the fix comment. `/home/brett/.claude/settings.json` carries **exactly one** `SessionStart` entry: matcher `startup|resume|clear|fork`, command `~/projects/xFactory/lanes-edit.sh session-start || true`, `"timeout": 5` — byte-identical to `claude-profile:41-43`, to `setup-claude-profiles.sh:214-216`, to `new-workstation#16` `c769cc2` **621**/**628**/**630**, and to Amendment 9 `#18` `fc33f39` **349**/**351**.

*Observation for Amendment 9's adoption, not a finding here:* `#18` `fc33f39` **393-399** budgets a migration on the premise that Eagle's hand-installed entry is `bash -lc '"$HOME/projects/xFactory/lanes-edit.sh" session-start 2>/dev/null || true'` and therefore makes `--install` refuse. As measured above, the installed entry **is** the canonical string. There is nothing left for that migration to remove.

### Opinion

**READY** — as a review opinion, not a ratification, and I am not marking this PR ready.

Every RV-W and RV-S finding is closed, including the one that was BLOCKING, and RV-S1's fix is closed on the row it was measured broken on rather than on a fixture. The three suites grew where the fixes landed (`EXPECTED_SCENARIOS` 4 → 9 in skill-install, with the shipped `SKILL.md` lines executed against fixture rows rather than a hand-copied stand-in), and both mutations I chose — one on this round's own fix, one a regression check on the prior round's — bit in the assertion written for them. The canonical hook string now agrees across five surfaces including the second writer this PR added. RV-S4 remains correctly inert: nothing here is in effect until `opensoft/brett-wip#4` lands.

**This is a review opinion, not a ratification.**

Issue #22, opened 2026-09-08
Shape-aware setup-openspeckit and Speckit worktree overlay (openRepoShape three-leg)


brettheap:
Lane: inkrouter-2 (InkRouter-2)

Governing record for the object **"setup-openspeckit and the Speckit worktree overlay become shape-aware (openRepoShape three-leg projects)"**, opened under lane-collision-protocol Rule 1 so the object has a place for claims.

Realization: #21 (branch `shape-aware-openspeckit`). First consumer: InkRouter/IRRS#5.

Contract: `~/.agents/protocols/project-agent-bootstrap.md` ("Worktree Layout", "Three-leg Projects", "The Speckit Worktree Overlay in a Three-leg Root") and `~/.agents/protocols/openspec-speckit-workflow.md` ("Repository Shape", "Worktree Rules").

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QbCk29uEjbed6TvPCoxFh9

brettheap:
CLAIMED — lane InkRouter-2, session harness 992a90f8-157a-43f4-8f86-658ca8d0b06c / PR-footer session_01QbCk29uEjbed6TvPCoxFh9 (earlier footer session_01PB38NpUuypicGg5TgyFQir) / session name InkRouter-2, 2026-09-08T22:52Z, for the shape-aware setup-openspeckit + Speckit worktree overlay (opensoft/workBenches#21).

(lane InkRouter-2, 2026-09-08T22:52Z) Retroactive: #21 was opened 2026-09-08T16:15Z before this window was named; the lane was registered at 2026-09-08T22:52Z.

Reads before claiming:
1. Comments on this record: none (issue opened by this lane; the repository had no issues).
2. `gh pr list --repo opensoft/workBenches --state all --search "openspeckit OR shape-aware"`: #21 OPEN (this lane); #13 MERGED (bootstrap portability, 2026-09-01, complete); #1 MERGED (cleanup).
3. `git ls-remote --heads origin | grep -i 'openspeckit\|shape'`: `shape-aware-openspeckit` only (this lane).
No live sibling claim, PR, or branch on this object.

brettheap:
[...]

PR #21, opened 2026-09-08
Make OpenSpec/Speckit bootstrap and worktree overlay shape-aware

brettheap:
Lane: inkrouter-2 (InkRouter-2)
Claim: https://github.com/opensoft/workBenches/issues/22#issuecomment-5592974013

## Summary

- `setup-openspeckit` detects an openRepoShape three-leg project from `project.yaml` (`--shape auto|on|off`), writes OpenSpec scaffolding into the spec leg and everything else at the assembly root, defaults `worktree_root` to `worktrees`, ignores `/worktrees/` in the root, appends a shape section to the managed agent block while preserving the `AGENTS-shape.md` first line, prefixes dry-run output with `[root]`/`[spec]`, and ends with the two-commit note.
- It upserts a managed `OPENSPEC-SPECKIT-SHAPE` block into every Speckit and OpenSpec skill and opsx command file (repo-local and user-global), refreshes overlays lacking the `speckit-overlay-shape` marker, replaces symlinked overlay skill directories with real copies, and embeds the current protocol texts.
- The worktree overlay's bash scripts gain `load_repo_shape` and a three-leg mode: the feature hook creates the `NNN-feature` branch in both legs from their fetched base, adds paired worktrees at `worktrees/<NNN>/{spec,code}`, numbers across both legs, writes the root's `.specify/feature.json`, and never touches the root repository. Discovery, the `ct` launchers, auto-commit, and initialize-repo follow. Single-repository behaviour is unchanged.
- opsx command and skill templates and the three overlay skills carry the shape block and no longer instruct agents to write code from OpenSpec tasks.

Contract: `~/.agents/protocols/openspec-speckit-workflow.md` ("Repository Shape", "Worktree Rules") and `~/.agents/protocols/project-agent-bootstrap.md` ("Worktree Layout", "The Speckit Worktree Overlay in a Three-leg Root").

## Test plan

- [x] `test-openspeckit-bootstrap.sh`: GREEN, 742 pass (new three-leg fixtures, `--shape`, first-line invariant, spec-leg refusals, family holder, shape blocks, overlay refresh, symlinked skill replacement)
- [x] `test-speckit-git-feature.sh`: GREEN, eight new three-leg scenarios plus all existing
- [x] `test-speckit-git-compat.sh` and `test-ct-launchers.sh`: GREEN
[...]

PR #67, opened 2026-09-12
devBenches/base-image: re-vendor openRepoTools at 0eb29b3 — `status` reads a parked record the way `resume` refuses it: no `parked_commit:`, no role, and whole-feature verdicts (#21, #22)

brettheap:
The `openrepotools` source pin moves from a40cbb7403fcc730c4590e8bd284196d0a91f9f0 (pinned in #59) to 0eb29b34daed39960c9ecf89c2600644a9fed784 — `gh api repos/opensoft/openRepoTools/compare/main...0eb29b34daed39960c9ecf89c2600644a9fed784` says `identical`, so it is the commit opensoft/openRepoTools's main carries after opensoft/openRepoTools#22.

Between the two commits, both squash-merged 2026-09-12: opensoft/openRepoTools#21 — `status`'s record layer reads a leg whose record names no `parked_commit:` as the RR1 refusal `resume.sh` makes ("has moved since it was parked") instead of offering `resume <Name>`, and a leg whose `- role:` value is empty as the whole-feature refusal `collect_legs` makes, naming the exit (park that feature again from the workstation that has it); `record_scalar` cuts a `#` comment before it unquotes, for every value `record_rows` reads. opensoft/openRepoTools#22 — the record is read PER FEATURE before any leg, the way `collect_legs` judges it: a feature `resume` refuses whole (a role this checkout's shape does not mount, no legs at all, a blank branch) never has a leg line offering `resume <Name>`; the absent-`pushed:` line names the refusal rather than one loader's reading; `record_rows` holds a block's rows until the key that selects the block is read, so a `root:` below the `features:` no longer loses the block or drops its legs; and a full-line `#` comment at column 0 inside a `project.yaml` leg no longer ends the leg.

Only `status` changes bytes: `update-upstream.py apply` reported `unchanged openRepoTools`, `unchanged park`, `unchanged resume`, `take status`. No consumer edits, because no command was added; update-upstream.py and the Dockerfile are untouched.

devBenches/devcontainer.test/test-speckit-git-feature.sh: 59 PASS, GREEN, exit 0.
The base image was not rebuilt here; the pin row and one vendored file are the only changes.

Lane: openrepotools-11 (openRepoTools-11)

## Summary by Sourcery

Re-vendor openRepoTools to incorporate corrected status handling for parked and partially configured feature records.

Bug Fixes:
- Update the vendored openRepoTools status command to correctly handle parked records, feature-level refusals, comments, and multi-feature record parsing.

Build:
- Re-vendor openRepoTools and update the pinned upstream commit and file checksum.

Tests:
- Validate the updated command behavior with the estate-command and speckit-git-feature test suites.

### DIFF
--- a/devBenches/base-image/files/openrepotools/status
+++ b/devBenches/base-image/files/openrepotools/status
@@ -145,14 +145,20 @@ pin that has fallen behind is named with the exit, `update-shape.py check
 workspace repository; `workspaces/<org>/<id>.yaml` in it) is read against the
 worktrees on disk: a recorded feature with no worktree here (`resume <Name>`
 brings it back — unless the record says `--no-push`, or its `pushed:` is
-missing or neither true nor false, which `resume` refuses the same way: then
-only the workstation that parked it can park it again; and where `git worktree
+missing or neither true nor false, which `resume` refuses the same way, or it
+names no parked commit for that leg, which `resume` refuses as moved-on, or it
+gives a leg no role at all, or a role this checkout's shape does not mount, or
+lists no leg for it at all, which `resume` refuses the whole feature for: then
+only the workstation that parked it can park it again — and where the role is
+really another shape's, only a checkout of that shape; and where `git worktree
 list` still holds a registration that is no longer a worktree, that is cleared
 with `worktree prune`, after a `worktree unlock` if it is locked and with any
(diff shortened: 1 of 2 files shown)


brettheap:
@codex review

PR #68, opened 2026-09-12
setup.sh: the openRepoTools wip init step (lane protocol Amendment 9)

brettheap:
Lane: openRepoProject-1

← brettheap/new-workstation#17 (Lane Collision Protocol **Amendment 9**, DRAFT). Built to the A9 text as spec: `protocols/lane-collision-protocol-amendment-9.md` on brettheap/new-workstation#18. **DRAFT — do not merge.** Amendment 9 is not ratified, and adoption act 3 (the openRepoTools PR) has not opened.

## What this does

The fourth link in Amendment 9(e)'s chain, in the order the amendment gives it (lines 351–361):

```text
gh repo clone opensoft/workBenches && cd workBenches && ./setup.sh   # the host, and the estate commands
openRepoTools --install                            # setup.sh already did this for you
openRepoTools wip init                             # <- THIS PR
pclaude run <profile> --lane <repo>-<n>
lane-start <repo> <n>
```

- **`scripts/setup-workspace-repo.sh`** (new) — the step.
- **`setup.sh`** — calls it under `log_header "WORKSPACE REPOSITORY"`, immediately after `scripts/setup-estate-commands.sh` and with nothing between them, best-effort exactly as the estate step is.
- **`devcontainer.test/test-setup-workspace-repo.sh`** (new) — 138 checks.
- **`README.md`** — the Quick Start list, the `setup.sh Flow` diagram, the re-run paragraph, and a new **Your Workspace Repository** section.

**Ordering is enforced three ways**, not just documented: the step consults `$OPENREPOTOOLS_BIN_DIR/openRepoTools` (the file `--install` just placed) before `$PATH`; it exits **2** naming `openRepoTools --install` and `scripts/setup-estate-commands.sh` when no openRepoTools is installed; and a static assertion in the suite fails if `setup.sh` ever calls it first or stops calling it.

## The prompt count is ZERO

Brett Heap's direction of 2026-09-12T18:56Z, verbatim: *"there are two options in resume. we need to also keep that clean so there is the least choices possible  to not confuse the user"*.

**A new person meets zero prompts in this step, in every path.** Nothing it can derive is asked:

| input | derived from | asked? |
|---|---|---|
| the login | `gh api user -q .login`, lowercased — A9(c) step 2, the naming contract's own definition of `<user>` | no |
| the organisation | the person's home organisation, `opensoft` in this estate — A9(c) step 3 | no |
| the repository | `<org>/<login>-wip`, from the two above | no |

The one-question budget the direction allows is **deliberately unspent**. Where nothing derives — no `gh`, or a `gh` that is not logged in — the step prints the one command that fixes it (`gh auth login`) and continues, rather than asking for what it could not read. Guarded two ways: statically, that the file contains no `read`/`select` at all; behaviourally, that the step completes with **stdin closed**.

Two `wip init` surfaces sit downstream of this step and are its own, not ours: ruling 2's print-and-continue wait for an administrator, and A9(c) step 4's exit 2 when stdin is not a terminal. Neither is a choice between options; both are named in the PR that implements them (act 3).

## The degradation contract

`openRepoTools` has **no `wip` subcommand today** — act 3 is sequenced after opensoft/brett-wip#4 merges — so this is the live path, not a hypothetical.

**Capability is read from `openRepoTools --help`, never from an exit code.** Today's openRepoTools answers an unknown argument through `die`, whose default exit is **2** — and 2 is *also* what A9(c) gives `wip init`'s own refusals (line 251–252). One number, two meanings. A probe that read the exit code would report a genuine refusal — with the administrator's `gh repo create` block already on the terminal — as "no such subcommand", and print the degradation recipe over the top of the one thing the person needed. Scenario (c3) fails if anyone rewrites the probe that way. Same shape as #63's `lane-start --help` probe for `--confirm`.

| situation | behaviour | exit |
|---|---|---|
| `~/.agents/workspace.yaml` names a repository and a checkout of it | prints what it found, runs **nothing** — not even `--help` | **0** |
| `wip` present, `wip init` succeeds | relays its output unchanged | **0** |
| **`wip` absent** | prints what the person would run (`update-upstream.py apply` + `./setup.sh`, or `openRepoTools wip init` directly), names the repository it would create, changes nothing | **0** |
| `wip init` refuses (missing rights, bad login, no tty) | **relays every byte it printed**, then a short trailer naming the exit code and that a re-run is safe (no summary of the block itself) | **1** |
| no openRepoTools at all (`--install` has not run) | refuses naming the step that must run first | **2** |
| `WORKBENCHES_SKIP_WORKSPACE_REPO=1` | one line | **0** |

`setup.sh` treats any non-zero here as best-effort and continues, as it does for the estate step above it.

**Nothing `wip init` prints is ever captured.** Its stdout and stderr are this script's, which are `setup.sh`'s, which are the terminal's. That matters most for the missing-rights path: Addendum 1 to #17 establishes the printed block is **three** administrator acts — `gh repo create`, the team-permission `PUT`, and the exclusion of `<login>-wip` from the organisation's PR-only ruleset, without which Rule 9's one-commit-per-write cannot land and the register is unwritable. Scenarios (d) and (d2) assert all three survive, through this step and through `setup.sh`'s own `|| echo` caller shape.

Idempotence is a file test on `~/.agents/workspace.yaml`, not a network call. Anything less than a repository plus a checkout of it falls **through** to `wip init`, which owns that judgement — A9(a) line 156: *"a second way to find it is a second answer"*. For the same reason `--login` is never passed to `wip init`; `--org` is, and only when `WORKBENCHES_WORKSPACE_ORG` is set, because an operator override has no other way in.

## Test run

Updated after closing re-verification review [#pullrequestreview-5188565322](https://github.com/opensoft/workBenches/pull/68#pullrequestreview-5188565322) (RV-W1–RV-W8); see that review and the closing comment on this PR for the finding-by-finding detail.

```
devcontainer.test/test-setup-workspace-repo.sh   138 checks, 0 failed
devcontainer.test/test-setup-estate-commands.sh  exit 0   (sibling, no regression)
devcontainer.test/test-claude-profile-lane-start.sh exit 0 (guard, untouched)
bash -n on setup.sh, the step and the suite            clean
```

Scenario-based and **mutation-checked**, in the style of #63. This closing round re-ran three of the earlier fix round's own reverting mutations (F1, F2, F3) and added one new mutation per RV-W1–RV-W4 blocking finding; all seven were caught against the current 138-check suite, and the unmutated baseline is clean:

| mutation | checks that failed |
|---|---|
| F1: the `--help` capability probe reverted to a bare-word `wip` match | 4 |
| F2: `workspace_repo_matches` forced to always return false | 17 |
| F3: the `<your-login>-wip` placeholder reintroduced | 3 |
| RV-W1: the `ssh://git@github.com/…` case arm removed from `normalize_github_repo` | 4 |
| RV-W2: the checkout-root (`rev-parse --show-toplevel`) comparison dropped from `workspace_repo_matches` | 2 |
| RV-W3: the `command -v timeout` guard dropped from `derive_login` | 3 |
| RV-W4: `cd workBenches &&` reverted back out of the clone example | 1 |

Coverage: subcommand present (a, i) · absent, against a fake **and against the real vendored `openRepoTools`** (c, c2) · the exit-code-probe mutation (c3) · idempotent re-run, three times over, byte-comparing `workspace.yaml` (b, b2) · a `path:` naming a subdirectory of the real checkout (b5, RV-W2) · every GitHub remote URL spelling — https/git@/ssh://, with and without `.git` (l, RV-W1) · the onboarding clone example is runnable as printed, in both the script header and README.md (m, RV-W4) · the missing-gh-rights path through the step and through `setup.sh`'s caller (d, d2) · ordering, behavioural and static (e1–e3) · the SETUP COMPLETE marker reminder cannot be skipped by the BUILD_FAILED exit path (e4, RV-W7) · the prompt count (f) · the skip switch (g) · the derivations (h1–h5) · the login still derives with no `timeout` binary on PATH (h6, RV-W3).

These suites are wired into CI. `.github/workflows/speckit-git-bash.yml` runs `devcontainer.test/test-setup-workspace-repo.sh` and the sibling `devcontainer.test/test-setup-estate-commands.sh` in the same job (run steps at `:80-84`), gated on `setup.sh`, `scripts/setup-workspace-repo.sh` and both test files in the `paths:` lists for `push` and `pull_request` (`:20-23`, `:41-44`). They also continue to run by hand and in the devcontainer, unchanged from how #37's suite is run today.

## Sequencing

**opensoft/workBenches#63's loop is the sole writer of the shared `lane-swap` skill for Amendment 8, today.** `scripts/setup-claude-profiles.sh:184-190` (`for skill in lane-swap; do … done`, at `fa00129`) installs `SKILL.md` into both `${CLAUDE_PROFILES_HOME:-$HOME/.claude-profiles}/shared/skills/lane-swap/SKILL.md` and `~/.claude/skills/lane-swap/SKILL.md`, from its own vendored copy — R-A8-5(a)/(c)'s two paths. This PR does not touch that loop, that directory, or either file.

**Amendment 9 act 4b — the pin-rows follow-up, sequenced after act 3 — deletes that loop**, together with the vendored copy it installs from, once `openRepoTools --install` owns both paths (act 3, clause (b)). It sits in 4b rather than earlier because deleting it before act 3 has merged would leave the skill with no writer at all — the same failure shape as a digest-less pin row, and it lands on every host the same way. Until act 4b lands, #63's loop remains authoritative for those two paths; this PR's `setup.sh` edit does not change that.

**The launcher's per-run `SessionStart` hook ensure is a different write and is not deleted.** `base-image/files/claude-profile` rewrites each profile's `settings.json` on every run and already preserves hook keys, so it ensures Amendment 8 clause (e)'s `SessionStart` entry there on every run, idempotently, by exact string match — **ratified R-A8-5(b)**. `openRepoTools --install` writes no profile's `settings.json` in any act of Amendment 9, so there is exactly one writer of that entry before, during and after the handover in 4b.

## Open questions

**Adoption act 4's sequencing and the `$AGENT_PROTOCOL_ROOT` pointer are ratified, not open — former Questions 1, 3 and 4 close here.** **R-A9-5** settles all three: the setup step in this PR may land at any time because it degrades by design, while the pin rows and the verify-list growth cannot open until act 3's commit exists on openRepoTools' default branch, because a pin row with no digest is refused structurally and that refusal takes the estate commands off every host (former Question 1, and former Question 3's naming of "the setup step" — no longer arguable from Addendum 1(1) alone, now from the ruling itself). The pointer file is `$AGENT_PROTOCOL_ROOT/workspace.yaml` — the name `resume` and `status` already use — and Amendment 9 mints no second one (former Question 4); this step's `${AGENT_PROTOCOL_ROOT:-$HOME/.agents}` is that name, not a guess act 3 might overrule. **R-A9-3** governs the one question below instead — team derivation is `wip init`'s act-3 judgement, not this step's — so Question 2 stays open on its own terms. Ratified 2026-09-12T22:26:38Z, against brettheap/new-workstation#18 at head `746715d`.

**2. `<team>` does not derive.** A9(c) step 4 writes `--team <team>` in both administrator commands and never says where the value comes from. It is `wip init`'s to resolve (act 3), not this step's — which is the second reason the degraded path prints the *pin move* and `openRepoTools wip init` rather than reproducing the create block: a placeholder the person must fill is a question wearing a different hat.

## Summary by Sourcery

Add a prompt-free, best-effort workspace repository initialization step to the setup flow after estate command installation, with graceful pre-adoption behavior and comprehensive validation.

New Features:
- Add a prompt-free workspace repository setup step that derives the GitHub login and organization and initializes the user's workspace repository when needed.
- Support graceful degradation until the installed openRepoTools provides `wip init`, including actionable setup guidance without modifying the host.
- Add workspace repository status tracking and completion reminders for setup failures or administrator-required actions.

Bug Fixes:
- Prevent workspace initialization from running before estate commands are installed and avoid invoking it when required GitHub prerequisites are unavailable.
- Preserve all output from `openRepoTools wip init`, including administrator instructions and refusal details.
- Ensure reruns are idempotent only for a matching repository checkout and correctly handle GitHub remote URL variants and nested checkout paths.

Enhancements:
- Integrate workspace repository initialization immediately after estate command installation in `setup.sh` as a best-effort step.
- Document the updated setup flow, workspace repository behavior, environment controls, and rerun semantics.

CI:
- Run the new workspace repository regression suite in the Git Bash CI workflow and trigger it for relevant source and test changes.

Documentation:
- Update the README setup checklist, flow diagram, rerun guidance, and workspace repository documentation.

Tests:
- Add comprehensive regression coverage for capability detection, ordering, idempotence, prerequisite handling, prompt-free execution, output propagation, skip behavior, and setup failure paths.


brettheap (commented):
Lane: openRepoProject-1

**Adversarial review of opensoft/workBenches#68 (DRAFT) @ `bbad3b6`.** A review opinion, not a ratification. Reviewed against the current head of the A9 text (`origin/lane-protocol/amendment-9-tooling-home`, brettheap/new-workstation#18), and against **A9 Addendum 2 (R-A9-1..7) and A8 Addendum 2 (R-A8-5), both ratified by Brett Heap** — so R-A9-3, R-A9-5 and R-A8-5 are read below as settled rules. Brett's 18:56Z direction on choices is read as the test it is named as.

**What I re-ran and confirmed, before the findings.** The suite: **76 checks, 0 failed**. Four of the eight mutations, re-derived independently — exit-code probe **20 failed**, capture-and-summarise **7**, drop the idempotence early return **3**, pass `--login` **2** — matching the PR's table on three of four (my capture mutation also folded stderr, hence 7 not 5). Baseline copy clean at 76/0. The zero-prompt claim holds **on a real pty**, not only with stdin closed: under `script -qec` (verified `stdin IS a tty`) the step ran to `EXIT=0` inside a 20s timeout, so nothing was waiting to be answered. The `--help` probe reads the real vendored shim correctly (`devBenches/base-image/files/openrepotools/openRepoTools --help` → 0 occurrences of `wip`, stdout-only). Login lowercasing works (`BrettHeap` → `brettheap`). The sibling suite is still green (`test-setup-estate-commands.sh` exit 0) and `bash -n` is clean on all three shell files. `WORKBENCHES_SKIP_WORKSPACE_REPO` matches the sibling's `= "1"` convention exactly. `AGENT_PROTOCOL_ROOT` is the right choice and complies with R-A9-5.

**Prompt count, since that is the direction's test.** In this step: **zero**, in every path, proven two ways here. In the `./setup.sh` run around it: the step is the third of eleven headers, and downstream of it a new person still meets `setup.sh:54` (Docker install y/N), `scripts/setup-claude-profiles.sh:67,71,78,46` (four, read from `/dev/tty`, so `</dev/null` does not silence them), `scripts/setup-workbenches.sh:162,299`, and the interactive TUI at `setup.sh:259`. The step spends none of its one-question budget; the chain around it is where the direction still has work, and that is not this PR's to fix.

Eleven findings, five blocking.

---

### F1 — BLOCKING. The capability probe matches a bare `wip` anywhere in `--help`, so prose claims capability

`scripts/setup-workspace-repo.sh:212`:

```sh
if ! printf '%s' "$tools_help" | grep -Eq '(^|[^[:alnum:]-])wip([^[:alnum:]-]|$)'; then
```

The hyphen guard is right — `<login>-wip` correctly does **not** match. A bare word in prose does. I built an `openRepoTools` with **no `wip` dispatch at all**, whose `--help` merely reads *"The helpers read your workspace repository, the wip repository your register lives in."* The step claimed capability and ran the verb:

```
Workspace repository: .../openRepoTools wip init
  target: opensoft/brettheap-wip (derived; you are asked nothing)

REFUSED: openRepoTools installs the estate commands and does nothing else.

Workspace repository step refused: `openRepoTools wip init` exited 2.
...Re-run `openRepoTools wip init` once that is done -- it is idempotent.
EXIT=1
```

That is the failure this PR's whole thesis exists to prevent, running in the other direction: a missing subcommand reported as a refusal, the `die` text handed to the person as if it were `wip init`'s own reasoning, and the one instruction that would actually help — move the pin — never printed. The step tells them to re-run the command that cannot exist. This is not hypothetical: A9(b) rewrites `openRepoTools:13` so the tool's identity line describes the workspace repository, and any help text that mentions it in prose before the verb is dispatchable trips this. The sibling probe this PR cites as its house style is tighter precisely because its token is distinctive — `claude-profile:393`, `[[ "$lane_start_help" == *"--confirm"* ]]`.

**Fix:** match the verb pair, not the word — `grep -Eq '(^|[[:space:]])wip[[:space:]]+init([[:space:]]|$)'` — and add a scenario to `devcontainer.test/test-setup-workspace-repo.sh` whose `FAKE_TOOLS_HELP` mentions `wip` in prose only and whose `wip` arm dies, asserting the step degrades (exit 0, "no `wip` subcommand yet") and never invokes the verb.

### F2 — BLOCKING. The idempotence detect is weaker than A9(c) step 1, and takes a judgement the file's own comment says it must not

`scripts/setup-workspace-repo.sh:153`:

```sh
if [ -n "$existing_repo" ] && [ -n "$existing_path" ] && [ -e "$existing_path/.git" ]; then
```

A9(c) step 1 is *"its `path:` is a checkout of that repository"*, and A9(a) makes the mismatch a **new** refusal in so many words: *"a `path:` that is not a checkout of the `repository:` it names beside it → exit 1 … nothing validates it today … so a `path:` pointing at some other repository is currently accepted and after this is not."* Two demonstrations against this branch:

- `repository: opensoft/brettheap-wip` with `path:` pointing at a checkout whose origin is `git@github.com:opensoft/workBenches.git` → `Workspace repository: opensoft/brettheap-wip already checked out at …` / `nothing to do.` / `EXIT=0`.
- `path:` pointing at a directory containing an **empty file** named `.git` → identical output, `EXIT=0`.

In both, `wip init` is never consulted, so the refusal A9(a) newly requires can never fire on this host — the step has silently answered a question it does not own, and it never re-asks, because the same file test passes forever. The comment at `:127-132` states the opposite as the design (*"Anything less — a half-written file, a path that is no checkout — falls THROUGH to `wip init`, which is idempotent and is the owner of that judgement"*); a `.git` file is not a checkout and it does not fall through. Scenario (b2) at `:230-242` only covers a path that does not exist.

**Fix:** gate the early return on both, still local, still no network:

```sh
if [ -n "$existing_repo" ] && [ -n "$existing_path" ] \
   && git -C "$existing_path" rev-parse --git-dir >/dev/null 2>&1 \
   && git -C "$existing_path" remote get-url origin 2>/dev/null \
      | grep -qi -- "$existing_repo\(\.git\)\?\$"; then
```

Anything else falls through to `wip init` as the comment promises. Add both scenarios above to the suite.

### F3 — BLOCKING. R-A9-3 forbids the placeholder this step prints on both paths, and the suite pins it

R-A9-3, ratified: **derive-or-omit, never a placeholder.** A9(c) step 4 gives the reason — *"A placeholder a person must fill is a question wearing a different hat"* — and this PR's own Open Question 2 quotes it back. `scripts/setup-workspace-repo.sh:200-204`:

```sh
if [ -n "$LOGIN" ]; then
    TARGET="$ORG/$LOGIN-wip"
else
    TARGET="$ORG/<your-login>-wip"
fi
```

Both branches print it. Degraded path, `gh` absent (`:234`):

```
  It will create opensoft/<your-login>-wip and ask you nothing.
```

And the **run** path, `gh` absent with a `wip`-capable tool (`:258`) — which no scenario covers at all:

```
  target: opensoft/<your-login>-wip (derived; you are asked nothing)
```

The second is worse than the first: it asserts *derived* about a value that was not derived. And `devcontainer.test/test-setup-workspace-repo.sh:493` and `:502` assert the placeholder as **correct** — *"the undeducible half is shown as a placeholder, not asked for"* — so the suite locks in the thing the ruling forbids.

**Fix:** delete the `else` branch at `:202-204` and leave `TARGET` empty. Where `TARGET` is empty, omit the `It will create …` sentence at `:234` and the `  target: …` line at `:258` entirely, keeping the existing `gh auth login` / `gh is not installed` note which already names the one command that fixes it. Flip `:493` and `:502` to `assert_not_contains "$OUTPUT" '<your-login>'`, and add the run-path-with-no-`gh` scenario asserting no `target:` line is printed.

### F4 — BLOCKING. "Not wired into CI" is false, and the sibling's runner is already in the file

The PR body states: *"These suites are not wired into CI — `.github/workflows/` carries no runner for `devcontainer.test/`, and the sibling estate-commands suite is not run there either."* On this branch, `.github/workflows/speckit-git-bash.yml:76-77`:

```yaml
      - name: Run the setup-estate-commands regression test
        run: bash ./devcontainer.test/test-setup-estate-commands.sh
```

with `scripts/setup-estate-commands.sh`, `devcontainer.test/test-setup-estate-commands.sh` and **`setup.sh`** in both `paths:` lists (`:20-22`, `:39-41`). So this PR's own `setup.sh` edit already triggers that workflow — which will run the sibling suite and not this one. The claim is not merely wrong, it describes the exact precedent the PR needed.

**Fix:** add `scripts/setup-workspace-repo.sh` and `devcontainer.test/test-setup-workspace-repo.sh` to both `paths:` lists, and a step after `:77`:

```yaml
      - name: Run the setup-workspace-repo regression test
        run: bash ./devcontainer.test/test-setup-workspace-repo.sh
```

Correct the PR body sentence. Note the consequence the suite already anticipates and CI makes load-bearing: scenario (c2) at `:263-280` asserts the real vendored shim *lacks* `wip`, so the act-3 pin-move PR must update (c2) in the same commit or land red. Say so in (c2)'s comment.

### F5 — BLOCKING (on the pair, #68 + #63). Two owners for the shared skills directory, and setup.sh's order makes workBenches win

R-A8-5, ratified, and inherited unvaried by A9(b): the skill is placed at `${CLAUDE_PROFILES_HOME:-$HOME/.claude-profiles}/shared/skills/lane-swap/SKILL.md` and the `~/.claude` pair kept for a bare run — and A9(b) says plainly *"those two are `--install`'s"*. But opensoft/workBenches#63 (`fa00129`) adds to `scripts/setup-claude-profiles.sh`:

```sh
for skill in lane-swap; do
  skill_source="$repo_dir/base-image/files/claude/skills/$skill/SKILL.md"
  [[ -f "$skill_source" ]] || continue
  mkdir -p "$base/shared/skills/$skill" "$default_claude_dir/skills/$skill"
  install -m 0644 "$skill_source" "$base/shared/skills/$skill/SKILL.md"
  install -m 0644 "$skill_source" "$default_claude_dir/skills/$skill/SKILL.md"
done
```

— the same two paths, from workBenches' own vendored copy. And the order in the file this PR edits decides who wins: `setup.sh:183` ESTATE COMMANDS (`openRepoTools --install`, `setup-estate-commands.sh:252`) runs **before** `setup.sh:220` AI PROFILE SETUP, so on every `./setup.sh` workBenches overwrites whatever `--install` placed, silently, whenever the two copies differ. That is the "second answer" defect A9 names three times, in the one directory R-A8-5 exists to make single. (The neighbouring `~/.claude/settings.json` write is fine — `setup-claude-profiles.sh:183-193` merges `.statusLine` onto the existing object with `jq` and preserves `hooks`, so A9's `SessionStart` entry survives. I checked; no finding there.)

**Fix:** this PR owns the A9 half of `setup.sh`, so record the conflict here and name the owner. When act 3 lands, act 4's pin follow-up deletes #63's `for skill in lane-swap` loop and the vendored `base-image/files/claude/skills/lane-swap/SKILL.md`, leaving `--install` sole owner per R-A8-5/A9(b); until then the README's Workspace Repository section and the PR body should say which write is authoritative. If instead workBenches is meant to keep it, that is an amendment to A9(b), not a silent race.

---

### F6 — minor. The README's numbered Quick Start contradicts `setup.sh` and the README's own tree

`README.md:16-17` lists the estate commands as **5** and the workspace repository as **6**, after VPN (2), Wave (3) and Docker (4). `setup.sh` runs them at `:183` and `:198` — immediately after SHELL SETUP (`:173`) and before VPN (`:214`) — which is what the tree this PR also edits says at `README.md:72-76`. The two halves of one file disagree, and the half a new person reads first is the wrong one.

**Fix:** move the two new items to positions 2 and 3 and renumber 4–9.

### F7 — minor. The degraded path promises a repository `wip init` is required to refuse

A9(c) step 2 makes a login that cannot form `^[a-z0-9]+(?:-[a-z0-9]+)*-wip$` *"a refusal naming the pattern and its source, never a name repaired by guesswork."* With `gh api user -q .login` returning `Brett_Heap`, this step prints at `:234`:

```
  It will create opensoft/brett_heap-wip and ask you nothing.
```

`brett_heap-wip` fails that pattern, so the command the step just told them to run will refuse. Not covered by any scenario.

**Fix:** with F3 applied, validate before forming `TARGET` — `case "$LOGIN" in *[!a-z0-9-]*|-*|*-|*--*) LOGIN="" ;; esac` — so a non-compliant login takes the same omit path as an underivable one, and add one line naming the pattern and openRepoShape's naming contract as its source. Add the scenario.

### F8 — minor. A network call on the path that does nothing, before the probe, with no bound

`:194` `gh api user -q .login` runs at `:198`, **before** the capability probe at `:211`, on every invocation that is not the idempotent early return. Today's live path is the degraded one, where the call's only consumer is the cosmetic `TARGET` string — so on every `./setup.sh` the step contacts GitHub in order to print a line and exit 0. There is no timeout anywhere in the file, and this is `setup.sh`'s third header, ahead of the VPN step `setup.sh:214` that the same script exists to configure: a host mid-VPN-setup or behind a captive portal blocks here with no message. (The sibling has no timeouts either, so this is a house gap rather than a regression — but this step is the one that puts a network call into the first minutes of `./setup.sh`.)

**Fix:** move `derive_login` / `ORG` / `TARGET` (`:191-204`) below the probe at `:211`, and bound the call: `login="$(timeout 10 gh api user -q .login 2>/dev/null || true)"`.

### F9 — minor. The administrator block reaches the pipe and not the screen

The relay is genuinely uncaptured — scenarios (d) and (d2) prove it, and I reproduced (d2)'s shape. But in the run that matters it is printed at header 3 of 11 and then buried under CLAUDE WORKFLOW, VPN, four `/dev/tty` prompts in AI PROFILE SETUP, the TUI at `setup.sh:259`, and Layer 0 / Layer 1 Docker builds; `setup.sh:302-312` SETUP COMPLETE then prints only `✓ Setup complete.` and the log path. The PR's strongest claim — three administrator acts must reach the person — is true of stdout and false of the terminal they are looking at when setup ends.

**Fix:** on exits 1 and 2, have the step `touch "$AGENTS_DIR/.workspace-step-needs-attention"`; in `setup.sh`'s SETUP COMPLETE block, print one line when that file exists, naming the step and `grep -n "WORKSPACE REPOSITORY" "$LOG_FILE"`, and remove the marker on a later successful run.

### F10 — minor. Mutation-harness hygiene, and two assertions weaker than they read

`devcontainer.test/test-setup-workspace-repo.sh:41` takes `$1` as `SCRIPT_UNDER_TEST` — a good affordance, and how I ran the mutations — but never checks it is executable. A mutant left at mode 644 reports **53 of 76 failed** (`exit 126` on every scenario), which reads exactly like a caught mutation; the real counts for the same four mutants are 20 / 7 / 3 / 2. Anyone re-deriving the PR's table by `sed` into a new file hits this, as I did. Separately, `:340` passes a count into `assert_contains`, so `1` also matches `10` or `21`.

**Fix:** after `:41`, `[ -x "$SCRIPT_UNDER_TEST" ] || { printf 'not executable: %s\n' "$SCRIPT_UNDER_TEST" >&2; exit 2; }`. Change `:340` to `assert_equal "$(grep -c 'continuing workBenches setup' <<<"$OUTPUT_D2")" '1' …`.

### F11 — minor. Two Open Questions are settled and should be closed, not carried

Brett Heap has ratified A9 Addendum 2 and A8 Addendum 2. **Open question 4** (`AGENT_PROTOCOL_ROOT` "left to the implementation") is settled by R-A9-5 and by the current #18 head — *"The directory is `$AGENT_PROTOCOL_ROOT`, and this text mints no second name for it"* — and the PR's choice is exactly right; it should be cited as compliance, not carried as risk. **Open questions 1 and 3** are settled by R-A9-5's *"act 4 is the setup step (opensoft/workBenches#68, may land first), THEN the pin rows and verify lists only after act 3's commit exists"*: this PR's scope is correct and no longer arguable from Addendum 1(1) alone, which it currently quotes.

**Fix:** replace both with one line citing R-A9-5 and R-A9-3.

---

### What the suite does not cover

A `--help` mentioning `wip` in prose only (F1); a `path:` that is a checkout of a different repository, and a `.git` that is a file (F2); the **run** path with no derivable login, where the placeholder is worst (F3); a login that cannot form a compliant name (F7); `--dry-run` on the degraded path; `WORKBENCHES_WORKSPACE_ORG` set but empty (behaviour is correct — both `:199` and `:261` treat it as unset — but unasserted); and the step under CI at all (F4).

### Opinion

**NOT READY** as a draft to make ready. The shape is right and unusually well argued — the ordering is enforced three ways, the relay is real, the idempotence early return does invoke nothing, and the zero-prompt claim survives a pty. F1 and F2 are correctness defects in the two mechanisms the PR rests on; F3 is a ratified ruling contradicted in code and asserted as correct in the suite; F4 is a false statement about the repository with a one-block fix; F5 is a settled-rule collision with #63 that the ordering in this PR's own `setup.sh` edit decides. None needs a redesign.

brettheap:
Lane: openRepoProject-1

**Shutdown notice — committing and stopping now, mid fix-round, on Brett Heap's word.** Head is now `5e40f12` (was `bbad3b6`), pushed to `setup/amendment-9-wip-init`.

Fix round on the adversarial review [pullrequestreview-5188323242](https://github.com/opensoft/workBenches/pull/68#pullrequestreview-5188323242) (F1–F11, F1–F5 blocking).

**Done and pushed, all verified green in this commit:**

| Finding | Status |
|---|---|
| F1 (blocking) | Fixed — probe now matches `wip[[:space:]]+init`, not a bare `wip`. Reproduction added (c4). |
| F2 (blocking) | Fixed — idempotence now requires `git rev-parse --is-inside-work-tree` + normalised `remote get-url origin` match, not `.git` existence. Comment at :127-132 fixed. Reproductions added (b3 foreign checkout, b4 `.git`-as-file). Also fixed the `yaml_value` quote/comment-order bug Copilot flagged inline on the same lines. |
| F3 (blocking, R-A9-3) | Fixed — no `<your-login>` placeholder anywhere. Run path now stops before calling a `wip init` it already knows would refuse (no/unauthenticated gh), exits 2. Tests h4/h5 flipped; new scenarios j1/j2 added. |
| F4 (blocking) | Code done — suite wired into `.github/workflows/speckit-git-bash.yml` (both `paths:` lists + a run step next to the sibling suite). **PR body/README "not wired into CI" claim NOT yet corrected — text edit only, remains open.** |
| F5 (blocking) | No code change needed — this PR's step is already independent of #63's `setup-claude-profiles.sh` loop; untouched. **PR body "Sequencing" ownership note NOT yet added — remains open.** |
| F6 | Fixed — README Quick Start renumbered to match `setup.sh`'s actual order. |
| F7 | Fixed — a login that can't form `<login>-wip` is treated as underivable, never guessed. Degraded-path (c5) and run-path (j3, stops before calling wip init) scenarios added. |
| F8 | Fixed — login derivation moved below the `--help` probe, `timeout 10` added. |
| F9 | Fixed — exits 1/2 leave `$AGENT_PROTOCOL_ROOT/.workspace-step-needs-attention`; `setup.sh` SETUP COMPLETE surfaces it; a clean run clears it (d3, j1). |
| F10 | Fixed — suite refuses a non-executable `SCRIPT_UNDER_TEST`; the d2 count assertion is now `assert_equal`, not `assert_contains`. |

**Remaining — not started, PR-body/documentation text only, no code:**
- F4: correct the false "not wired into CI" sentence in the PR body.
- F5: add the "Sequencing" note to the PR body (act 4's pin-rows half also deletes #63's loop).
- F11: consolidate Open Questions 1, 3, 4 into one line citing R-A9-5 and R-A9-3; Question 2 (`<team>` does not derive) stays open, untouched.
- Final PR comment with the full finding→commit table and before/after counts, as the task asked for, was not reached.

**Verification on `5e40f12`, all green:**
- `devcontainer.test/test-setup-workspace-repo.sh`: 76 → **110 checks, 0 failed**.
- Sibling suites unchanged and green: `test-setup-estate-commands.sh` (exit 0), `test-claude-profile-lane-start.sh` (exit 0).
- `bash -n` clean on `setup.sh`, `scripts/setup-workspace-repo.sh`, `devcontainer.test/test-setup-workspace-repo.sh`.
- Mutation set: the original 8 re-derived and still caught (counts shifted upward with the larger suite, e.g. exit-code-probe now 28/110 not 20/76), plus 5 new mutations for this round's fixes (F1 bare-word revert, F2 `.git`-existence-only revert, F3 placeholder revert, F7 pattern-check revert, F9 marker-removal revert) — **13/13 caught**, unmutated baseline clean at 110/0.

Not marked ready; not merged. Next session should pick up at the PR-body edit (F4/F5/F11 text) and the final finding→commit table comment.

brettheap:
Lane: openRepoProject-1

**Fix round on review [pullrequestreview-5188323242](https://github.com/opensoft/workBenches/pull/68#pullrequestreview-5188323242) closed out.** Progress comment [issuecomment-5649228264](https://github.com/opensoft/workBenches/pull/68#issuecomment-5649228264) left three text-only items pending after `5e40f12`; they are now applied as a PR-body edit (F4's CI-claim correction, a new **Sequencing** section for F5, and the **Open Questions** consolidation for F11). Full finding → closure map, F1–F11:

| Finding | Closed by | What closed it |
|---|---|---|
| F1 (blocking) | `5e40f12` | Capability probe now matches the verb pair `wip[[:space:]]+init`, never a bare `wip` in `--help` prose. Reproduction added (c4). |
| F2 (blocking) | `5e40f12` | Idempotence detect now requires `git rev-parse --is-inside-work-tree` **and** a normalised `remote get-url origin` match to `workspace.yaml`'s `repository:`, not mere `.git` existence. Also fixed the `yaml_value` quote/comment-order bug (review 5187716029) on the same lines. Reproductions added: foreign checkout (b3), `.git`-as-file (b4). |
| F3 (blocking, R-A9-3) | `5e40f12` | No `<your-login>` placeholder on either path. Run path now stops before invoking a `wip init` it already knows would refuse, exits 2. Tests h4/h5 flipped to `assert_not_contains`; scenarios j1/j2 added. |
| F4 (blocking) | `5e40f12` (CI wiring) **+ body edit** (claim correction) | Code: `scripts/setup-workspace-repo.sh` + its suite added to both `paths:` lists and a run step added to `.github/workflows/speckit-git-bash.yml`. Text: the PR body's false "not wired into CI" sentence is corrected to cite the workflow's run steps and `paths:` lists by line number. |
| F5 (blocking) | body edit only, no code | No code change was needed — this PR's step is already independent of opensoft/workBenches#63's `setup-claude-profiles.sh` loop. A new **Sequencing** section now records: #63's loop is the sole writer of the shared `lane-swap` skill for Amendment 8 today; Amendment 9 act 4b (after act 3) deletes that loop once `openRepoTools --install` owns both paths; the launcher's per-run `SessionStart` hook ensure is a separate write and is **not** deleted (ratified R-A8-5(b)). |
| F6 | `5e40f12` | README Quick Start renumbered so estate commands (2) and workspace repository (3) match `setup.sh`'s actual order. |
| F7 | `5e40f12` | A login that cannot form `<login>-wip` is treated as underivable, never guessed. Degraded-path (c5) and run-path (j3) scenarios added. |
| F8 | `5e40f12` | Login derivation moved below the `--help` probe; bounded with `timeout 10`. |
| F9 | `5e40f12` | Exits 1/2 leave `$AGENT_PROTOCOL_ROOT/.workspace-step-needs-attention`; `setup.sh`'s SETUP COMPLETE surfaces it; a later clean run clears it (d3, j1). |
| F10 | `5e40f12` | Suite refuses a non-executable `SCRIPT_UNDER_TEST` instead of misreading exit-126 as caught mutations; the d2 count assertion changed to `assert_equal`. |
| F11 | body edit only, no code | Open Questions consolidated: former Questions 1, 3 and 4 close under one paragraph citing **R-A9-5**; Question 2 (`<team>` does not derive) stays open, untouched, per **R-A9-3** governing that derivation as act 3's (`wip init`'s), not this step's. |

**Verification, on `5e40f12`:**
- `devcontainer.test/test-setup-workspace-repo.sh`: **76 → 110 checks**, 0 failed.
- Mutation set: the original 8 re-derived and still caught, plus 5 new mutations for this round's fixes (F1, F2, F3, F7, F9) — **13/13 caught**, unmutated baseline clean.
- Sibling suites unchanged and green (`test-setup-estate-commands.sh`, `test-claude-profile-lane-start.sh`); `bash -n` clean on all touched shell files.

Head: `5e40f12b66aec5eba3776486f17c2ae1425c278c`. Still DRAFT, not marked ready, not merged.

brettheap (commented):
Lane: openRepoProject-1

**Checklist RE-VERIFICATION of the fix round at `5e40f12`** (review 5188323242, F1-F11), against the current #18 text and the ratified rulings. Re-run and independently re-derived, not taken from the PR's own account.

## Confirmed: the suite, and F1-F4/F6-F10 by direct code read

`bash ./devcontainer.test/test-setup-workspace-repo.sh` → **110 checks, 0 failed**, matching expectation exactly. Three reverting mutations run against the fixed code, each restored after:
- F1's regex (`(^|[[:space:]])wip[[:space:]]+init([[:space:]]|$)`) reverted to the old bare-word match → **4 failed**.
- F2's idempotence gate (`workspace_repo_matches`, git-checkout + normalized-origin comparison) forced false → **5 failed**.
- F3's placeholder removal (`TARGET=""`) reverted to reintroduce `<your-login>-wip` → **3 failed**.

All three mutations caught, confirming the suite guards the fixes.

Verified directly in `scripts/setup-workspace-repo.sh`: F1's exact fixed regex at `:258`; F2's `workspace_repo_matches()`/`normalize_github_repo()` at `:190-208`, reproduced against both of F2's original demonstrations (a foreign-origin checkout, and an empty file named `.git`) — both now correctly fall through rather than short-circuiting; F3's `TARGET=""` with no `else` branch (`:310-312`), and zero occurrences of `<your-login>` anywhere outside the test file's `assert_not_contains` checks; F6 (README numbering, estate commands at 2 / workspace repo at 3, matching `setup.sh`'s order); F7 (login-pattern validation, `^[a-z0-9]+(-[a-z0-9]+)*$`); F8 (`timeout 10 gh api user -q .login` moved after the capability probe — see RV-W3 below, though, for a regression this introduces); F9 (needs-attention marker, confirmed set/cleared correctly in the suite); F10 (`[ -x "$SCRIPT_UNDER_TEST" ]` guard added, count assertions switched to `grep -c` + `assert_equal`).

Zero-prompt claim independently reproduced two ways: stdin closed (`</dev/null`, exit 0) and on a real pty (`script -qec`, confirmed `stdin IS a tty`, exit 0).

F4's CI wiring confirmed directly: `scripts/setup-workspace-repo.sh` and `devcontainer.test/test-setup-workspace-repo.sh` are in both `paths:` lists of `.github/workflows/speckit-git-bash.yml`, and a dedicated `Run the setup-workspace-repo regression test` step exists.

F5 (skill-loop collision with #63) and F11 (Open-question consolidation) are correctly out of scope for this PR's code — F5's actual fix is Amendment 9 act 4b (`:958-966` in the #18 text, confirmed present), and F11 is a PR-body edit. Per the brief, judged on code only; both listed as **pending**, not failing.

## Remaining findings — 4 new BLOCKING, found by independent reproduction against the current code

**RV-W1 — BLOCKING. `normalize_github_repo()` does not handle the `ssh://git@github.com/<repo>` URL form.** `:190-203` only cases on `https://github.com/*` and `git@github.com:*`; the comment's own "one of the four spellings" is short one. Reproduced: a checkout with `origin` set to `ssh://git@github.com/opensoft/brettheap-wip.git` and a matching `workspace.yaml` still falls through past the idempotence check instead of matching — the exact contract F2 just fixed, defeated by an unhandled fourth spelling. **Fix:** add a third case, `ssh://git@github.com/*) url="${url#ssh://git@github.com/}" ;;`.

**RV-W2 — BLOCKING. The idempotence check accepts a `path:` that is a subdirectory of the real checkout, not its root.** `git -C "$path" rev-parse --is-inside-work-tree` and `remote get-url origin` both succeed from inside any subdirectory of a work tree. Reproduced: `path:` pointed at `<checkout>/nested/` (not the checkout root) still reported `already checked out at .../nested — nothing to do`. Since Amendment 9(a) resolves the register at `<path>/lanes/LANES.md`, a subdirectory `path:` silently resolves to the wrong file downstream. (A Copilot review comment on this PR attributes the fix pattern to a "shared validator" in `workspace-common.sh:293-299` — checked, that file is `devBenches/base-image/files/speckit-worktree/templates/specify/extensions/git/scripts/bash/workspace-common.sh`, an unrelated speckit-worktree template; the citation is wrong, but the underlying defect is real and independently reproduced here.) **Fix:** additionally require `[ "$(git -C "$path" rev-parse --show-toplevel)" = "$(cd "$path" && pwd -P)" ]` (or equivalent) before accepting the match.

**RV-W3 — BLOCKING, and it is a regression introduced by this round's own F8 fix. `timeout 10 gh api user -q .login` (`:277`) is unguarded, and this repository already has the guarded pattern it should have followed.** `scripts/ensure-layer3.sh:184-185` checks `command -v timeout` before use — because GNU `timeout` is not on a stock macOS PATH (exactly the platform F8's fix was motivated by: "common on macOS" per the reviewing bot). Reproduced: with `timeout` absent from `PATH`, the expression `"$(timeout 10 gh api user -q .login 2>/dev/null || true)"` silently returns empty — `|| true` swallows the "command not found" — so a fully-authenticated `gh` is misreported as unauthenticated and the step wrongly tells the person to run `gh auth login`. **Fix:** guard it — `if command -v timeout >/dev/null 2>&1; then login="$(timeout 10 gh api user -q .login 2>/dev/null || true)"; else login="$(gh api user -q .login 2>/dev/null || true)"; fi` — matching `ensure-layer3.sh`'s own precedent.

**RV-W4 — BLOCKING, shared with brettheap/new-workstation#18 (see that PR's RV-T2). The onboarding example is not runnable as printed.** `README.md:97` and this script's own header comment (`:8`) both print `gh repo clone opensoft/workBenches && ./setup.sh`; `gh repo clone` cannot change the caller's cwd, so `./setup.sh` is looked up in the original directory and fails. Flagged (unresolved, "Suppressed comments") by Copilot review 5188460935 on both lines. Reproduced directly (see #18's review for the transcript). **Fix:** `cd workBenches &&` in both places.

## Minor / lower-severity, also independently confirmed

**RV-W5 — minor. The degradation contract's own documented table is contradicted by the code.** The PR body's table states a `wip init` refusal "relays every byte it printed, adds no summary." `scripts/setup-workspace-repo.sh:419-425` prints four extra lines after the relay (`Workspace repository step refused: ...`, etc.) before `mark_needs_attention` and `exit 1`. Either trim those four lines or correct the documented contract.

**RV-W6 — minor, confirmed. The recovery hint points at a log file that was never written.** `setup.sh:198-202` invokes the workspace step directly (not through `run_logged`), so its stdout/stderr go only to the terminal — `log_header` (`setup.sh:20-24`) writes only header/timestamp lines to `$LOG_FILE`. But the SETUP COMPLETE reminder (`setup.sh:322`) tells the person to `grep -n "WORKSPACE REPOSITORY" "$LOG_FILE"`, which will find only the empty header block, never the administrator block or refusal text they need. **Fix:** call the step through `run_logged` (or tee both streams), matching the pattern that helper already provides.

**RV-W7 — minor, confirmed by inspection. The marker reminder can be skipped entirely.** `setup.sh:305-308` exits 1 when `BUILD_FAILED=true`, which is checked *before* the workspace-repo marker block at `:313-323`. A host with a first-run Docker/Layer-1 build failure never sees the reminder, even though the marker file exists. **Fix:** move the marker check above the `BUILD_FAILED` exit, or check it from an exit trap.

**RV-W8 — minor, pending not failing (concurrent PR-body edit in progress per the brief).** The PR body's "Test run" section still reads "76 checks" and the pre-fix 8-mutation table; the current suite is 110 checks, and my independent 3-mutation re-run produced different counts than the stale table. Also flagged by Copilot review 5188460935. Update the body once the concurrent edit lands.

## Opinion

**NOT READY.** F1-F4 and F6-F10 are correctly and verifiably fixed, and the suite is solid (110/0, mutation-tested). But RV-W1-W3 are gaps in the exact two mechanisms (idempotence, capability/derivation) this round just touched, found by direct reproduction rather than by trusting the fixed code's own account of itself — R-A9-3's "derive-or-omit, never a placeholder" is undermined by RV-W3 in one specific, real environment (no GNU `timeout`). RV-W4 is shared with #18 and should be fixed in both PRs together.

Review opinion only; this lane ratifies nothing.

brettheap:
Lane: openRepoProject-1

Closing the re-verification findings from [review #pullrequestreview-5188565322](https://github.com/opensoft/workBenches/pull/68#pullrequestreview-5188565322) (RV-W1–RV-W4 BLOCKING, RV-W5–RV-W8 minor).

## Finding → commit

| finding | fix | commit |
|---|---|---|
| RV-W1 (BLOCKING) | `normalize_github_repo()` now cases on `ssh://git@github.com/<repo>(.git)?` as a third spelling, alongside `https://` and `git@…:` | `d8aea7e` |
| RV-W2 (BLOCKING) | `workspace_repo_matches()` now also requires `git -C "$path" rev-parse --show-toplevel` to equal `$path` itself, both realpath-normalised (`pwd -P`) — a `path:` naming a subdirectory of the real checkout no longer reads as "already done" | `d8aea7e` |
| RV-W3 (BLOCKING) | `derive_login()` now checks `command -v timeout` before using it, falling back to a bare `gh api user -q .login` call when absent (stock macOS), matching the guard already in `scripts/ensure-layer3.sh` | `4130570` |
| RV-W4 (BLOCKING) | The onboarding clone example in this script's own header and in `README.md:97` now reads `gh repo clone opensoft/workBenches && cd workBenches && ./setup.sh`; the PR body's own copy of the same example was corrected too | `5ab1ad0` (+ body edit) |
| RV-W5 (minor) | Corrected the PR body's degradation-contract table: the refusal path relays every byte of `wip init`'s own output verbatim, then adds a short trailer naming the exit code (not a summary of the block itself) — code already matched this; only the doc wording was wrong | body edit only, no code change |
| RV-W6 (minor) | `setup.sh` now invokes the step through `run_logged "WORKSPACE REPOSITORY" ...`, so the step's own stdout/stderr are teed into `$LOG_FILE`, not just the header banner — the SETUP COMPLETE recovery hint now points at a log slice that actually has content | `2040d32` |
| RV-W7 (minor) | The `.workspace-step-needs-attention` marker check in `setup.sh`'s SETUP COMPLETE summary now runs *before* the `BUILD_FAILED` exit, so a first-run Layer 1 build failure can no longer skip it | `2040d32` |
| RV-W8 (minor) | PR body test counts and mutation table updated to the final numbers below | body edit only, no code change |

## Tests

Each commit above lands green on its own (re-verified in isolation, not just at the end):

| stage | checks | sibling suites | `bash -n` | mutations caught |
|---|---|---|---|---|
| baseline (`5e40f12`) | 110, 0 failed | green | clean | — |
| after `d8aea7e` (RV-W1, RV-W2) | 131, 0 failed | — | clean | — |
| after `4130570` (RV-W3) | 135, 0 failed | — | clean | — |
| after `5ab1ad0` (RV-W4) | 137, 0 failed | — | clean | — |
| after `2040d32` (RV-W6, RV-W7) — final | **138, 0 failed** | `test-setup-estate-commands.sh` exit 0; `test-claude-profile-lane-start.sh` exit 0 | clean on all touched files | **7/7 caught** |

Mutation set at the final head (`2040d32`): the three historical reverting mutations this review's own re-verification exercised (F1 bare-word `wip` match → 4 failed; F2 `workspace_repo_matches` forced false → 17 failed against the larger suite; F3 `<your-login>-wip` placeholder reintroduced → 3 failed) plus one new mutation per blocking fix (RV-W1 ssh:// case removed → 4 failed; RV-W2 toplevel check removed → 2 failed; RV-W3 timeout guard removed → 3 failed; RV-W4 clone-example `cd` reverted → 1 failed). All seven caught; unmutated baseline at `2040d32` is clean (138/0).

Counts before → after: **110 → 138 checks, 0 failed** at both ends; **+28 new checks** (RV-W1: 18 across six URL-spelling scenarios; RV-W2: 3; RV-W3: 4; RV-W4: 2; RV-W7: 1; RV-W6 modified an existing static assertion rather than adding one).

Head sha: `2040d321876fed9bbbf0b23547e0fc7681f7db2a`

brettheap (commented):
Lane: openRepoProject-1

## Final confirmation pass — RV-W1..W8 re-verification fix round, checked at `2040d32`

A checklist re-check against the RV-W1–W8 review (pullrequestreview-5188565322) and the fix comment (5649508761) — confirming closure at the current head, not re-reviewing from scratch. Opinion only, not a ratification.

| Finding | Severity | Status | Evidence |
|---|---|---|---|
| RV-W1 (`ssh://` URL form unhandled) | BLOCKING | CLOSED | `scripts/setup-workspace-repo.sh` `normalize_github_repo()`: `ssh://git@github.com/*) url="${url#ssh://git@github.com/}" ;;` added, with an inline comment explicitly labeled "RV-W1" naming the exact defect it closes. Read directly. |
| RV-W2 (subdirectory `path:` false match) | BLOCKING | CLOSED | `workspace_repo_matches()` now compares `git -C "$path" rev-parse --show-toplevel` against the realpath-normalized `$path` itself, rejecting a subdirectory match. |
| RV-W3 (`timeout` unguarded — regression from F8) | BLOCKING | CLOSED | `derive_login()` now branches on `command -v timeout`, with an inline comment explicitly labeled "RV-W3" citing the exact platform regression it fixes and the `scripts/ensure-layer3.sh:184-185` precedent it follows. Read directly. |
| RV-W4 (broken clone-then-setup example, shared w/ #18 RV-T2) | BLOCKING | CLOSED | Both `scripts/setup-workspace-repo.sh`'s header and `README.md:97` now read `gh repo clone opensoft/workBenches && cd workBenches && ./setup.sh`. |
| RV-W5 (degradation-table doc/code mismatch) | minor | CLOSED | Doc-only fix; PR body's degradation table now matches the actual refusal-block behavior (a trailer after the relay). |
| RV-W6 (recovery hint pointed at a log never written) | minor | CLOSED | `setup.sh` now routes the workspace step through `run_logged`, which genuinely tees stdout+stderr via `tee -a`. |
| RV-W7 (marker reminder reachable only on some exit paths) | minor | CLOSED | `setup.sh`: the needs-attention marker check now precedes the `BUILD_FAILED` exit. |
| RV-W8 (stale PR-body check-count) | minor | CLOSED | PR body consistently says 138 (3 places); no stale "76" remains. |

**Test suite — independently re-run twice (once via sub-verification, once directly by me):**
`devcontainer.test/test-setup-workspace-repo.sh` → **138 checks, 0 failed**, both times, exact match to the brief and the PR body. 13 further `devcontainer.test/*.sh` sibling suites and 3 of 4 `devBenches/devcontainer.test/*.sh` suites also pass; `test-ct-launchers.sh` fails on a missing `/usr/local/share/ct/ct-functions.zsh` — confirmed pre-existing/environment-only (this PR's commit touches nothing ct-related, the file is base-image-installed and absent from this host entirely, and that test file's own last modification was an unrelated PR #13). Not a regression from this PR.

**Mutation spot-checks (both requested mutations performed, both restored — worktree confirmed byte-identical afterward via `git status --short`/`git diff`, both empty, verified directly by me):**
- Removed the `ssh://` case from `normalize_github_repo()` → 4 checks failed, matching the PR body's own claimed mutation-row count for RV-W1 exactly. Restored cleanly.
- Removed the `command -v timeout` guard from `derive_login()` → 3 checks failed (the no-timeout-binary checks specifically), matching the PR body's RV-W3 mutation-row count exactly, and reproducing the precise misbehavior RV-W3 describes (a fully-authenticated `gh` misread as unauthenticated). Restored cleanly.

**PR-body vs. reality:** every count and degradation-table row checked matches current behavior; no mismatches found.

**Regression scan:** `git show 2040d32 --stat` = 2 files changed (+47/-11): `setup.sh` and the test suite only. Diff scope is exactly RV-W6 (`run_logged` wiring) and RV-W7 (marker-check reorder + a new assertion), matching the commit subject. No scope creep.

**Opinion: READY.** All eight RV findings are closed, verified directly against source rather than the fix comment's account of itself. The primary suite is genuinely green (re-run twice independently), both requested mutation spot-checks prove the suite actually catches the regressions it claims to guard against, and the one failing suite in a wider sweep is a pre-existing environment gap unrelated to this PR. No open items.

PR #27, opened 2026-09-09
Add a bounded WSL transport watchdog

brettheap:
Lane: workbenches-sync
Claim: https://github.com/opensoft/workBenches/issues/23#issuecomment-5593664129

## Summary

- add a bounded Windows-side WSL process-creation watchdog that kills only its own timed-out probe
- classify health after consecutive failures and capture bounded host/guest transition evidence
- encode child PowerShell commands so paths cannot alter argument boundaries
- provide explicit install, status, start, stop, and uninstall actions; uninstall removes the installed executable while preserving diagnostics
- keep WSL service restart, distribution termination, and shutdown outside the watchdog authority

## Validation

- Windows PowerShell AST parsing passed for both scripts
- two bounded failed probes against a nonexistent test distribution produced transition evidence without leaving child probes
- encoded guest snapshot completed successfully from a temporary state path
- Status emitted parseable JSON separately from exit code 3 when no daemon identity was active
- strict OpenSpec validation and git diff checks passed
- the existing scheduled task was not installed, started, stopped, or uninstalled during source validation

## Summary by Sourcery

Add a bounded Windows WSL transport watchdog that detects process-creation failures, records actionable evidence, and offers explicit non-disruptive lifecycle management.

New Features:
- Add a Windows-side WSL transport watchdog that performs deadline-bound process-creation probes and classifies consecutive failures.
- Add bounded host, guest, and Windows-event evidence capture when transport health becomes stuck.
- Provide explicit per-user scheduled-task installation and lifecycle actions for installing, checking, starting, stopping, and uninstalling the watchdog.

Enhancements:
- Preserve unrelated WSL workloads by limiting termination to watchdog-owned probe and diagnostic child processes and excluding service, distribution, and relay recovery actions.
- Sanitize child PowerShell invocation and diagnostic state while adding atomic status writes, bounded event rotation, and evidence retention.
- Document transport failure boundaries, operational state, diagnosis, and controlled recovery limitations.

Documentation:
- Document watchdog operation, runtime files, evidence, failure boundaries, and recovery guidance.

Tests:
- Validate PowerShell parsing, bounded probe behavior, status output and exit-code handling, encoded child arguments, diagnostic deadlines, and evidence retention.

PR #26, opened 2026-09-09
Refresh shared AI CLI tooling

brettheap:
Lane: workbenches-sync
Claim: https://github.com/opensoft/workBenches/issues/23#issuecomment-5593664129

## Summary
- refresh the shared AI CLI inventory and compatibility adapters
- pin npm 12.0.2 and fail closed unless selective lifecycle-script support is available
- make download pipelines propagate upstream failures and qualify only supported noninteractive call providers

## Validation
- no-cache shared base-image build
- disposable runtime probe for every required AI CLI and npm 12.0.2
- shell syntax, adapter profile, shared MCP, Claude tmux, and Codex profile tests
- strict OpenSpec validation
- no running bench container recreation

## Summary by Sourcery

Refresh the shared AI CLI baseline, profile integrations, and noninteractive routing while enforcing safer installation and authentication boundaries.

New Features:
- Expand shared images with additional AI coding CLIs and provider-specific profile recognition.

Bug Fixes:
- Propagate failures from piped native installer downloads and refresh the Herdr installer checksum.
- Preserve existing Claude profile model selections while using the updated default for new profiles.
- Prevent ambiguous Cursor command ownership and reconcile the shared NotebookLM MCP launcher.

Enhancements:
- Pin npm 12.0.2 and fail closed when selective lifecycle-script controls are unavailable.
- Restrict unified noninteractive provider routing to qualified Codex, Claude, and Gemini integrations while retaining broader inventory detection.
- Improve authentication detection for Qwen, Kimi Code, DeepSeek Harness, and MiniMax profiles.
- Document the expanded CLI inventory, authentication workflows, and routing boundaries.

Build:
- Refresh the shared image marker and update the system Google Cloud package to google-cloud-cli.

Documentation:
- Add operator guidance for Kimi Code, Qwen Code, MiniMax Code, DeepSeek Harness, and the Z.AI coding helper.

Tests:
- Add adapter routing, optional-install summary, and Claude profile preservation tests.

Chores:
- Add OpenSpec change documentation covering the refreshed shared AI CLI baseline and activation boundaries.

PR #24, opened 2026-09-09
Add a fail-closed WSL relay watchdog

brettheap:
Lane: workbenches-sync
Claim: https://github.com/opensoft/workBenches/issues/23#issuecomment-5593664129

## Summary

- add a fail-closed Linux-side watchdog for strict orphan WSL relay processes
- preserve named relays, SessionLeaders, relays with children on any thread, recent relays, malformed names, and ambiguous identities
- revalidate PID and start-time identity before TERM and before KILL
- refuse occupied unmanaged install paths and modified managed files using recorded SHA-256 digests
- provide reversible installation, bounded logging, dry-run inspection, and documentation

## Validation

- 26 focused unit tests passed
- Python compilation and shell syntax checks passed
- strict OpenSpec validation passed
- live non-mutating qualification with the same 60-second minimum-age policy found zero strict candidates, protected all 25 named relays and 24 SessionLeaders, and found zero scan errors or classification overlap
- git diff --check passed

## Summary by Sourcery

Add a fail-closed WSL relay watchdog that safely detects confirmed orphan relays and provides non-destructive lifecycle management.

New Features:
- Add a fail-closed WSL relay watchdog that identifies and cleans up only confirmed strict orphan relays while preserving named relays, SessionLeaders, child-bearing processes, recent processes, and ambiguous identities.
- Provide dry-run scanning, daemon status reporting, bounded logging, configurable cleanup limits, and reversible installation and rollback without enabling systemd or restarting WSL.

Enhancements:
- Require repeated PID/start-time identity confirmation and full predicate revalidation before TERM and KILL, using pidfds to prevent PID-reuse signaling.
- Make installation non-destructive by preserving unrelated WSL settings, refusing boot-command and managed-path conflicts, and tracking managed-file digests for safe upgrades and removal.

Deployment:
- Register and launch the watchdog through the WSL boot command with root-owned runtime, configuration, state, and log files.

Documentation:
- Document watchdog safety boundaries, installation, activation, status and scan commands, configuration bounds, conflict handling, and rollback procedures.

Tests:
- Add focused synthetic procfs, classification, revalidation, signaling, configuration, daemon, logging, installation, and rollback coverage, along with live non-mutating qualification and syntax checks.

PR #78, opened 2026-09-14
devBenches/base-image: re-vendor openRepoTools at 8a36eb3 and vendor its whole install set — `--install` places all eleven files, two skills, one command file and the hook or none (#26), so the four-file subset could no longer install offline (#23, #27, #26, #33)

brettheap:
Decision flagged for Brett Heap: vendoring all fourteen paths is the only way the offline install stays green under #26's all-or-nothing rule.

The offline install refused everything against the old four-file vendor set
(openRepoTools, park, resume, status only) once openRepoTools's own
`--install` grew to expect all eleven bin-directory files beside itself:

REFUSED: could not fetch restart lanes lanes-edit.sh lane-start lane-end link-estates repos.tsv from pinned-by-workBenches-no-fetch …

Decision (flagged for Brett Heap in the PR): vendor the whole install set,
not just the four bin files this repository already tracked. Upstream's
`--install` places all eleven bin-directory files, both skills and the one
command file, or NONE at all (opensoft/openRepoTools#26's all-or-nothing
rule: `collect_commands`, `collect_skills` and the `SessionStart` hook merge
each refuse rather than partially install). workBenches' own D3 guarantee —
no network is ever reachable during this host's install, enforced by the
OPENREPOTOOLS_REPO/_REF sentinel scripts/setup-estate-commands.sh exports —
means every one of those fourteen paths this repository does not already
hold beside itself must be vendored here, not fetched at install time. A
partial-install mode upstream would reverse a ratified rule (Amendment 11),
so vendoring the whole set is the only way the offline install stays green
under #26's rule.

Pin moved 0eb29b34daed39960c9ecf89c2600644a9fed784 ->
8a36eb344935444e36ff5af5551e5bf9157f7f15. `gh api
repos/opensoft/openRepoTools/compare/main...8a36eb344935444e36ff5af5551e5bf9157f7f15
--jq .status` = identical (re-checked at commit time): 8a36eb3 is main.

Five PRs land in that range (`git -C openRepoTools log --oneline
--first-parent 0eb29b34..8a36eb3`), in merge order:

#23 -- "`status` reads a parked feature the way `resume` refuses it before
any commit is compared — a branch origin has lost is RR2's refusal of the
WHOLE feature, and a role this checkout's shape has no place for is
`collect_legs`'s — reading this checkout's shape as `load_repo_shape` reads
it, with one parser of `project.yaml`". A `status` refinement; no bin-file or
install-surface change.

#24 -- "Lane tooling moves home: filter-repo from brett-wip, openRepoTools
wip init, --install installs the lane skill and hook (Amendment 9 act 3)".
THE CONSEQUENTIAL ONE: this is where `wip init`, the two skills, the command
file and the hook merge first shipped, and where `--install`'s installable
set first grew past the four files this repository tracked before today.
Landing this pin is what turns scripts/setup-workspace-repo.sh's own
`--help` probe live -- see below.

#26 -- "Lane tooling: restart, /restart, lanes, session-lane, window-lane,
dir and profile in the record (Amendment 11 tooling)". Adds `restart`,
`lanes` and the four lane helpers with `repos.tsv` to the installable set --
the bulk of the eleven-file growth this PR vendors.

#27 -- "`status` judges a parked feature's worktree by the PATH `resume`
computes for it — a worktree on that branch anywhere else and anything else
AT that path are both refusals of the WHOLE feature, asked in `resume.sh`'s
own order and stopped at the first leg that refuses — computing that path
out of the root GIT would spell, and out of `$SPECKIT_GIT_WORKTREE_ROOT`
before any file". A `status` refinement; no install-surface change.

#33 -- "`status` counts only the features and legs `workspace_load_project`
will make, and a `- branch:` or a `- role:` the record's KEY ORDER leaves
under a shut list is a finding naming the list that shut it and the two
exits in the order that does not lose the work, instead of a `resume <Name>`
that brings none of it back". The pin target itself; a `status` refinement,
no further install-surface change past #26's.

`update-upstream.py apply --source openrepotools --at
8a36eb344935444e36ff5af5551e5bf9157f7f15 --yes --add restart --add lanes
--add lanes-edit.sh --add lane-start --add lane-end --add link-estates --add
repos.tsv --add skills/lane-swap/SKILL.md --add skills/restart/SKILL.md
--add commands/swap.md` took all fourteen openrepotools paths (the ten new
ones added, the four already-pinned ones re-taken at the new commit).
`update-upstream.py check` exits 0: 16 vendored copies match (14
openrepotools + 2 openreposhape), across both sources.

Suite counts: devcontainer.test/test-setup-estate-commands.sh 182 PASS, 0
FAIL, exit 0. devBenches/devcontainer.test/test-speckit-git-feature.sh 59
PASS, 0 FAIL, exit 0. devcontainer.test/test-setup-workspace-repo.sh 152
checks, 0 failed, exit 0 (a new suite touched, below). Logs under
/workspace/projects/.lane-worktrees/openRepoTools-3/revendor-scratch/. The
Docker-based CI jobs (git-2-34-1, bash-3-2), which run the speckit-git
feature and compat suites under older git and bash, were not run locally.

devBenches/base-image/Dockerfile: COPY and chmod 0755 all eleven bin files,
not four (mirroring the installer, which stamps every one of the eleven
executable, `repos.tsv` included, regardless of its own pin-recorded mode);
the two skills, the command file and the hook stay per-user and are NOT
COPY'd into the image, and the comment block says so and says where they are
placed instead (scripts/setup-estate-commands.sh, on the host). The stale
"`park`, `resume` and `status` are vendored from files/openrepotools/"
paragraph -- which still named only those three after an earlier pass added
a second paragraph beside it -- is rewritten into one paragraph that carries
the history and names the eleven.

scripts/setup-estate-commands.sh: a `jq` preflight (the hook merge needs it)
before either shim runs, so a host with no `jq` refuses before anything is
staged rather than after ten of eleven bin files land; `require_pin_row` for
all fourteen openrepotools paths, not five, because #26's all-or-nothing rule
means any one missing a row refuses the whole install; `check_target_preflight`
and `verify_installed` for the eleven bin targets.

devcontainer.test/test-setup-estate-commands.sh: rewritten for the new
contract. Scenario (a) covers all eighteen artifacts (eleven bin files, both
skills at their two paths each, the command file at its two paths, the
merged hook) under a sandboxed $HOME -- now with CLAUDE_PROFILES_HOME and
CLAUDE_USER_DIR pinned explicitly alongside $HOME in every scenario, not
just relying on $HOME: those are the two variables `skills_home()`/
`claude_home()` actually read, and an ambient export of either (plausible on
a real workstation or CI runner) would otherwise redirect the skill/command/
hook install away from the sandbox regardless of $HOME. (b)/(c) cover
rerun-unchanged and a modified `park` replaced; (l) checks mode 0755 on all
eleven bin files; (k)'s D3 guarantee keeps its no-network shape in a first
part (this script's own `require_pin_row` refuses a documented row-and-file
removal before either shim runs, naming the file, with no "fetch" or
"github" ever appearing in the output) and gained a second part that calls
the vendored shim directly, with a file missing beside it, to prove the
sentinel backstop itself -- not just the front-door preflight -- refuses
loudly and names the sentinel, installing nothing. The `assert_mode` crash
(a missing file's `stat` failing under `set -e` and killing the whole
script, rather than failing one assertion) is fixed.

The workspace-repo step goes live: with `wip init` vendored, `setup.sh` now
runs `openRepoTools wip init` on the host when a person runs setup;
scenario (c2) updated as its own text prescribed; the Docker-based CI jobs
(git-2-34-1, bash-3-2) were not run locally. In full:
scripts/setup-workspace-repo.sh's own capability probe (`grep` for the verb
pair `wip init` in `--help`, never an exit code) now finds it in the real
vendored `--help`, so this step takes the live path instead of the degrade
path for the first time. devcontainer.test/test-setup-workspace-repo.sh's
scenario (c2) predicted this exact day in its own header comment ("the day
it does land, this scenario is the one that notices... THAT DAY, the act-3
pin-move PR must update this scenario... in the same commit, or CI (F4)
lands red") and is updated exactly as that text prescribed: it now asserts
the real vendored copy is seen to CARRY `wip init`, that the step invokes
it, and -- under a fake `gh` that grants nothing (no real network is ever
reachable: `env -i` puts the fake first on $PATH) -- asserts what `wip init`
asked the fake (`api user`, `api --paginate /user/teams`, `repo view`,
`repo create`, and never `repo clone`, since the fake never lets the
repository exist), the administrator block it prints with every value
substituted, and its own sandboxed refusal (exit 2, relayed as this step's
own exit 1) verbatim -- never forced to a fabricated "pass" with a `--yes`
or an auto-create path this command does not have. scripts/setup-workspace-repo.sh's
header comment (the "has no `wip` subcommand at all... so for now"
paragraph) was stale the same way and is reworded: the verb arrived with
this pin, the `--help` probe is what turns the step live on every run, and
the degradation path stays for an earlier or hand-installed copy.

setup.sh:184 runs this on the host, best-effort.

The base image was not rebuilt here.

Lane: openrepotools-3 (openRepoTools-3)


brettheap:
@codex review

